PR #5, opened 2019-08-18
[Testnet] Add Airdrop Contract

mrtpain:
Requesting to merge and whitelist the Airdrop contract.

**Description**
Generic Airdrop contract that can be used to automate distribution to registrants. Owners of the airdrop can specify a future endblock to close on or choose to manually close the registration period of the contract. Users register by sending their address, when registration is closed, users can call the contracts withdraw method to retrieve their tokens. The airdrop contract calls the tokens contract that is being distributed to transfer the allotted amount. The airdrop contract must be authorized to transfer up to the total supply being distributed.

**Compiler**
Compiled with SCT (commit e651c5)

**Hash**
```
0ac2a13a055e6f6f0b0e5bc3430406562af174e50a105334fe3f0abeeb03e1fc
```

**ByteCode**
```
4D5A90000300000004000000FFFF0000B800000000000000400000000000000000000000000000000000000000000000000000000000000000000000800000000E1FBA0E00B409CD21B8014CCD21546869732070726F6772616D2063616E6E6F742062652072756E20696E20444F53206D6F64652E0D0D0A2400000000000000504500004C010200407B8DA50000000000000000E00022200B0130000010000000020000000000004A2F0000002000000040000000000010002000000002000004000000000000000400000000000000006000000002000000000000030040850000100000100000000010000010000000000000100000000000000000000000F82E00004F000000000000000000000000000000000000000000000000000000004000000C000000DC2E00001C0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000200000080000000000000000000000082000004800000000000000000000002E74657874000000500F0000002000000010000000020000000000000000000000000000200000602E72656C6F6300000C0000000040000000020000001200000000000000000000000000004000004200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002C2F0000000000004800000002000500D4230000080B00000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000FE0203280500000A0205166AFE037201000070280600000A0205280500000602042803000006020E0428070000060202280700000A6F0800000A28090000062A4602280900000A721F0000706F0A00000A2A4A02280900000A721F000070036F0B00000A2A4602280900000A72490000706F0C00000A2A4A02280900000A7249000070036F0D00000A2A4602280900000A72610000706F0C00000A2A4A02280900000A7261000070036F0D00000A2A4602280900000A72730000706F0A00000A2A4A02280900000A7273000070036F0B00000A2A4602280900000A727F0000706F0C00000A2A4A02280900000A727F000070036F0D00000A2A4602280900000A72A70000706F0C00000A2A4A02280900000A72A7000070036F0D00000A2A7202280900000A72CD000070038C07000001280E00000A6F0F00000A2A7602280900000A72CD000070038C07000001280E00000A046F1000000A2A7A0228060000062C1402281100000A6F1200000A022806000006FE052A172A4A0202280700000A6F0800000A28160000062A8A02280700000A6F0800000A022808000006281300000A2C08020328160000062A162A0000133002003D0000000100001102280900000A72A70000706F0C00000A0A062D270228100000062D1F02280A000006166A361502280400000602280A0000065C0A0206280D000006062AB602280700000A6F0800000A022808000006281400000A2C02162A0202281100000A6F1200000A2807000006172A0013300600CB000000020000110202280700000A6F0800000A280E00000672E3000070281500000A2D080228100000062C02162A0228130000060A198D0800000125160228080000068C07000001A2251702280700000A6F0800000A8C07000001A22518068C10000001A20B020228020000060672F500007007166A281600000A0C02086F1700000A7201000070280600000A0202280700000A6F0800000A720F010070280F000006021203FE1503000002120302280700000A6F0800000A7D030000041203720F0100707D0400000409280100002B172A0013300300760000000300001103022808000006281300000A2C02162A0203280E000006281900000A2C160228100000062C0E02280A0000060228040000063702162A0202280A000006176AD7280B000006020372E3000070280F000006021200FE15030000021200037D03000004120072E30000707D0400000406280100002B172A000042534A4201000100000000000C00000076342E302E33303331390000000005006C0000001C040000237E0000880400001804000023537472696E677300000000A00800002001000023555300C0090000100000002347554944000000D00900003801000023426C6F620000000000000002000001571DA201090A000000FA01330016000001000000100000000300000004000000160000000F0000001900000002000000040000000300000001000000070000000D000000010000000200000001000000010000000000CA0101000000000006000B01500206003A0150020600F7003D020F00700200000A008303F0020A006700F0020A00E002F00206009103DC010A009803F00206005D00DC010A002B01F0020A004800F0020A008E00F00206008001DC010A009101F00206000E00DC01000000001500000000000100010001001000F50100001500010001000A0110008701000029000300170051804403C60051803703C6000600AB03C90006006E03C60050200000000086183702CD00010090200000000086088B0221000500A220000000008108A402D7000500B520000000008608D10359000600C720000000008108E103DD000600DA20000000008608A20159000700EC20000000008108AF01DD000700FF2000000000860813022100080011210000000081081D02D7000800242100000000860807035900090036210000000081081F03DD0009004921000000008108B30059000A005B21000000008108CA00DD000A006E210000000086005303E2000B008B210000000081006403E8000C00A921000000008608270285000E00C8210000000086002E0285000E00DB21000000008600A803EF000E000022000000008600E10059000F004922000000008600E30185000F007822000000008600C80385000F0050230000000081005801EF000F00000001007B0000000200BD0200000300F10300000400BC01000001006D01000001006D01000001006D01000001006D01000001006D01000001006D0100000100E80200000100E80200000200750300000100B60300000100B603090037020100110037020600190037020A005900370206002900370210002900C1031600290051001C0061000802210029009F0026006900D2022B006900DD02310069000100380069000B003D0071007C03430069007301490069007D014E002900980154007900FD0159003900FD035D00390009045D007100090473002900C501790049007F02850029008D018900710035009A000E000400A8000E000800B9002E000B0002012E0013000B012E001B002A01610023003301650069009500020001000000A802F5000000E503FA000000B301FA0000002102F50000002303FA000000E400FA0000002B02FE0002000200030001000300030002000400050001000500050002000600070001000700070002000800090001000900090002000A000B0001000B000B0002000C000D0001000D000D00020010000F00048000000000000000000000000000000000830300000400000000000000000000009F001E00000000000100020001000000000000000000F002000000000300020031009000000000000047657455496E7436340053657455496E743634003C4D6F64756C653E0053797374656D2E507269766174652E436F72654C69620049734E756C6C4F725768697465537061636500494D657373616765006765745F4D6573736167650056616C7565547970650049536D617274436F6E7472616374537461746500736D617274436F6E74726163745374617465004950657273697374656E745374617465006765745F50657273697374656E745374617465006765745F416D6F756E74546F44697374726962757465007365745F416D6F756E74546F4469737472696275746500476574416D6F756E74546F446973747269627574650044656275676761626C6541747472696275746500436F6D70696C6174696F6E52656C61786174696F6E7341747472696275746500496E6465784174747269627574650052756E74696D65436F6D7061746962696C6974794174747269627574650041646452656769737472616E74457865637574650076616C756500476574537472696E6700536574537472696E67005374617475734C6F670049426C6F636B006765745F426C6F636B006765745F456E64426C6F636B007365745F456E64426C6F636B00656E64426C6F636B0043616C6C00536D617274436F6E74726163742E646C6C0053797374656D00436C6F7365526567697374726174696F6E0041697264726F70006765745F4E756D626572006765745F53656E646572006765745F4F776E6572007365745F4F776E6572006765745F43616E5265676973746572002E63746F720053797374656D2E446961676E6F73746963730053797374656D2E52756E74696D652E436F6D70696C6572536572766963657300446562756767696E674D6F646573006765745F53756363657373006765745F546F6B656E436F6E747261637441646472657373007365745F546F6B656E436F6E74726163744164647265737300746F6B656E436F6E74726163744164647265737300476574416464726573730053657441646472657373006164647265737300537472617469732E536D617274436F6E747261637473006765745F4E756D6265724F6652656769737472616E7473007365745F4E756D6265724F6652656769737472616E74730046756E64656453746174757300456E726F6C6C6564537461747573004765744163636F756E74537461747573005365744163636F756E745374617475730073746174757300466F726D617400536D617274436F6E7472616374004F626A65637400495472616E73666572526573756C740041646452656769737472616E740072656769737472616E7400417373657274005769746864726177006765745F546F74616C537570706C79007365745F546F74616C537570706C7900746F74616C537570706C79006F705F457175616C697479006F705F496E657175616C6974790000001D41007300730065007200740020006600610069006C00650064002E00002954006F006B0065006E0043006F006E00740072006100630074004100640064007200650073007300001754006F00740061006C0053007500700070006C007900001145006E00640042006C006F0063006B00000B4F0077006E006500720000274E0075006D006200650072004F006600520065006700690073007400720061006E0074007300002541006D006F0075006E00740054006F00440069007300740072006900620075007400650000155300740061007400750073003A007B0030007D00001145004E0052004F004C004C004500440000195400720061006E007300660065007200460072006F006D00000D460055004E00440045004400000000003703AD195EBD5C4C8344A0AC9B6BBC740004200101080320000105200101111105200101121905200201020E0420001231042000111D0420001235052001111D0E062002010E111D0420010B0E052002010E0B0500020E0E1C0420010E0E052002010E0E042000123D0320000B07000202111D111D0307010B0907040B1D1C1225110C050002020E0E0B20051225111D0B0E1D1C0B0320000206300101011E00040A01110C040701110C040001020E087CEC85D7BEA7798E1045004E0052004F004C004C00450044000C460055004E0044004500440002060E0306111D092004011219111D0B0B05200101111D042001010B0520010E111D06200201111D0E05200102111D042800111D0328000B032800020801000800000000001E01000100540216577261704E6F6E457863657074696F6E5468726F777301080100020000000000040100000000000000000000000000000010000000000000000000000000000000202F000000000000000000003A2F00000020000000000000000000000000000000000000000000002C2F0000000000000000000000005F436F72446C6C4D61696E006D73636F7265652E646C6C0000000000FF25002000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002000000C0000004C3F00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
```


mrtpain:
One question I've been mulling over is using `TransferTo` or `TransferFrom` methods to transfer the tokens. 

Using `TransferTo` will require the contract address to hold the amount to be airdropped. 
Using `TransferFrom` will require the contract address to be approved to send the amount to be airdropped.

Preferences/Suggestions on which one would be better?

King-Pharaoh:
> One question I've been mulling over is using `TransferTo` or `TransferFrom` methods to transfer the tokens.
> 
> Using `TransferTo` will require the contract address to hold the amount to be airdropped.
> Using `TransferFrom` will require the contract address to be approved to send the amount to be airdropped.
> 
> Preferences/Suggestions on which one would be better?

I see your first implementation is _TransferTo_ which is what I probably would've opted for as well. Do you know what would happen if you used _TransferFrom_ with an amount larger than what was in the Contract Address? (disclaimer, looking at this purely from the surface and haven't dived into how these methods are defined 😄 )

mrtpain:
@King-Pharaoh I believe the same thing would happen if `TransferTo` tries to send more then it has in it's balance. Failed transaction and gas is lost. 

The differences between these are subtle but I think the contract address holding the funds would be better than the address being approved to move the funds from the owners wallet.

King-Pharaoh:
Agreed. There's an added layer of security separating from the main wallet to the contract wallet that I didn't consider. 

TransferTo does seem to be the better move.

Testnet/Airdrop/Airdrop.Tests/AirdropTests.cs
```diff
@@ -0,0 +1,651 @@
...
+        public void CloseRegistration_Success()
+        {
+            this.mockContractState.Setup(m => m.Message).Returns(new Message(this.airdropContract, this.airdropContractOwner, 0));
+            this.mockContractState.Setup(b => b.Block.Number).Returns(0);
+            this.mockPersistentState.Setup(e => e.GetUInt64("EndBlock")).Returns(0);
+            this.mockPersistentState.Setup(e => e.GetAddress("Owner")).Returns(this.airdropContractOwner);
+
+            // Initialize the smart contract set constructor props
+            var airdrop = new Airdrop(this.mockContractState.Object, this.tokenContract, this.totalSupply, this.endBlock);
+
```
Neurosploit: Assert registrationclosed is false in between

Testnet/Airdrop/Airdrop.Tests/AirdropTests.cs
```diff
@@ -0,0 +1,651 @@
...
+        [Fact]
+        public void CloseRegistration_Fail_SenderIsNotOwner()
+        {
+            this.mockContractState.Setup(m => m.Message).Returns(new Message(this.airdropContract, this.registrant, 0));
+            this.mockContractState.Setup(b => b.Block.Number).Returns(0);
+            this.mockPersistentState.Setup(e => e.GetUInt64("EndBlock")).Returns(this.endBlock);
+
+            // Initialize the smart contract set constructor props
+            var airdrop = new Airdrop(this.mockContractState.Object, this.tokenContract, this.totalSupply, this.endBlock);
+
```
Neurosploit: Same here

Testnet/Airdrop/Airdrop.Tests/AirdropTests.cs
```diff
@@ -0,0 +1,651 @@
...
+            var registrationIsClosed = airdrop.RegistrationIsClosed;
+
+            Assert.True(registrationIsClosed);
+
+        }
+
+        [Fact]
+        public void RegistrationIsClosed_IsFalse_IfCurrentBlockIsLessThanEndBlock()
```
Neurosploit: Add test for same block as end block.

Testnet/Airdrop/Airdrop.Tests/AirdropTests.cs
```diff
@@ -0,0 +1,651 @@
...
+
+        }
+        #endregion
+
+        #region Amount To Distribute Tests
+
+        [Fact]
+        public void AmountToDistribute_ReturnsCorrectAmount()
+        {
+            this.mockContractState.Setup(m => m.Message).Returns(new Message(this.airdropContract, this.registrant, 0));
```
Neurosploit: maybe also test 10 /4 to prove that it doesn't pay out too much.

Testnet/Airdrop/Airdrop/Airdrop.cs
```diff
@@ -0,0 +1,197 @@
...
+
+    /// <summary>Sets the status for a given address.</summary>
+    private void SetAccountStatus(Address address, Status status)
+    {
+        PersistentState.SetStruct($"Status:{address}", status);
+    }
+
+    /// <summary>Validates and registers accounts. See <see cref="AddRegistrantExecute(Address)"/></summary>
+    public bool Register()
+    {
```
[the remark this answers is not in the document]
mrtpain: This functionality was to provide a way for contract owners to sign up their own participants. For example, when a user signs up for an online account they automatically get enrolled. I also removed the ability for the owner to register themselves. Good call on that.

Testnet/Airdrop/Airdrop.Tests/AirdropTests.cs
```diff
@@ -0,0 +1,651 @@
...
+
+            for (uint i = 1; i < 5; i++)
+            {
+                this.mockPersistentState.Setup(s => s.GetUInt64("NumberOfRegistrants")).Returns(i);
+                Assert.Equal(i, airdrop.NumberOfRegistrants);
+            }
+        }
+
+        [Fact]
+        public void AddRegistrant_Success()
```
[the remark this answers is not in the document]
mrtpain: They both share the same private Execute method. I think only the minor difference should be checked, that the owner is the sender and the owner is not registering themselves.

Testnet/Airdrop/Airdrop.Tests/AirdropTests.cs
```diff
@@ -0,0 +1,651 @@
...
+            var _totalSupply = airdrop.TotalSupply;
+            Assert.Equal((ulong)1, _totalSupply);
+
+            var numRegistrants = airdrop.NumberOfRegistrants;
+            Assert.Equal((ulong)10, numRegistrants);
+        }
+
+        [Fact]
+        public void Register_Success_NumberOfRegistrantsIncrement()
+        {
```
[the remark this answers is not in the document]
mrtpain: Will work on this, it's hard to prove something didn't happen with tests.

Testnet/Airdrop/Airdrop.Tests/AirdropTests.cs
```diff
@@ -0,0 +1,651 @@
...
+                ExpectedAmountToDistribute = this.totalSupply,
+                transferResult = new TransferResult(success: true)
+            };
+            result = WithdrawExecute(withdrawParams);
+            Assert.False(result);
+        }
+
+        [Fact]
+        public void Withdraw_Fail_TokenContractTransferFailure()
+        {
```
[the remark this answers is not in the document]
mrtpain: I added a couple more scenarios to `AmountToDistribute_ReturnsCorrectAmount` which is what is used within the withdraw methods.

Testnet/Airdrop/Airdrop.Tests/AirdropTests.cs
```diff
@@ -0,0 +1,785 @@
...
+            this.mockPersistentState.Verify(s => s.SetUInt64("NumberOfRegistrants", airdrop.NumberOfRegistrants + 1));
+
+            for (uint i = 1; i < 5; i++)
+            {
+                this.mockPersistentState.Setup(s => s.GetUInt64("NumberOfRegistrants")).Returns(i);
+                Assert.Equal(i, airdrop.NumberOfRegistrants);
+            }
+        }
+
+        //[Fact]
```
Neurosploit:
```suggestion
          [Fact]
        public void Register_Fail_NumberOfRegistrantsNotIncremented()
        {
            ulong totalNumberOfRegistrants = 0;
            this.mockContractState.Setup(m => m.Message).Returns(new Message(this.airdropContract, this.airdropContractOwner, 0));
            this.mockContractState.Setup(b => b.Block.Number).Returns(999_999);
            this.mockPersistentState.Setup(e => e.GetUInt64("NumberOfRegistrants")).Returns(() => { return totalNumberOfRegistrants; });
            this.mockPersistentState.Setup(e => e.SetUInt64("NumberOfRegistrants", It.IsAny<ulong>())).Callback<string, ulong>((key, value) => { totalNumberOfRegistrants = value; });
            this.mockPersistentState.Setup(e => e.GetUInt64("EndBlock")).Returns(this.endBlock);
            this.mockPersistentState.Setup(e => e.GetUInt64("TotalSupply")).Returns(this.totalSupply);

            Dictionary<string, Status> registrations = new Dictionary<string, Status>();
            this.mockPersistentState.Setup(e => e.GetStruct<Status>($"Status:{this.registrant}"))
                .Returns<string>((key) =>
                {
                    if (registrations.ContainsKey(key))
                    {
                        return registrations[key];
                    }

                    return Status.NOT_ENROLLED;
                });
            this.mockPersistentState.Setup(e => e.SetStruct($"Status:{this.registrant}", It.IsAny<Status>()))
                .Callback<string, Status>((key, value) =>
                {
                    if (registrations.ContainsKey(key))
                    {
                        registrations[key] = value;
                    }
                    else
                    {
                        registrations.Add(key, value);
                    }
                });

            // Initialize the smart contract set constructor props
            var airdrop = new Airdrop(this.mockContractState.Object, this.tokenContract, this.totalSupply, this.endBlock);

            Assert.Equal((ulong)0, totalNumberOfRegistrants);
            this.mockContractState.Setup(m => m.Message).Returns(new Message(this.airdropContract, this.registrant, 0));

            // call first time with registrant that is not registered yet.
            var result = airdrop.Register();

            Assert.True(result);
            this.mockPersistentState.Verify(s => s.SetUInt64("NumberOfRegistrants", It.IsAny<ulong>()), Times.Once);
            Assert.Equal((ulong)1, totalNumberOfRegistrants);

            // call second time with same registrant that is already registered.
            result = airdrop.Register();
            Assert.False(result);
            Assert.Equal((ulong)1, totalNumberOfRegistrants);

            // verify it's not called more times than before. It should still be only once.
            this.mockPersistentState.Verify(s => s.SetUInt64("NumberOfRegistrants", It.IsAny<ulong>()), Times.Once);
        }
```

I suggest we start setting up these tests with a mocked storage behind them, see how I've done it. That will give more realistic test results as they now depend on a lot of 'static' mocking.

Testnet/Airdrop/Airdrop/Airdrop.cs
```diff
@@ -0,0 +1,202 @@
...
+
+        SetAccountStatus(Message.Sender, Status.FUNDED);
+
+        Log(new StatusLog { Registrant = Message.Sender, Status = Status.FUNDED });
+
+        return true;
+    }
+
+    /// <summary>Validates and adds a new registrant. Updates the NumberOfRegistrants, account status, and logs result.</summary>
+    private bool AddRegistrantExecute(Address registrant)
```
Neurosploit:
```suggestion
     /// <summary>Validates and adds a new registrant. Updates the NumberOfRegistrants, account status, and logs result.</summary>
    private bool AddRegistrantExecute(Address registrant)
    {
        if (registrant == Owner)
        {
            return false;
        }

        var accountStatus = GetAccountStatus(registrant);
        bool invalidAddressStatus = accountStatus != Status.NOT_ENROLLED;
        if (invalidAddressStatus || RegistrationIsClosed || NumberOfRegistrants >= this.TotalSupply)
        {
            return false;
        }

        // only add registrant if not enrolled yet.
        if (accountStatus == Status.NOT_ENROLLED)
        {
            NumberOfRegistrants++;

            SetAccountStatus(registrant, Status.ENROLLED);

            Log(new StatusLog { Registrant = registrant, Status = Status.ENROLLED });

            return true;
        }

        return false;
    }
```

This code has a bug, number of registrants is always incremented when the execute is called and the address status is valid while it already can be enrolled. The suggested test proves it. Here is the code suggestion to fix it,

mrtpain:
@Neurosploit committed your test, thanks for that. I'll refactor existing tests to match what you've done.

codingupastorm:
Hey @mrtpain, awesome. We'll have a good look at it now!

Testnet/Airdrop/Airdrop/Airdrop.cs
```diff
@@ -0,0 +1,202 @@
...
+    /// <summary>Returns the status of any given address.</summary>
+    public Status GetAccountStatus(Address address)
+    {
+        return PersistentState.GetStruct<Status>($"Status:{address}");
+    }
+
+    /// <summary>Sets the status for a given address.</summary>
+    private void SetAccountStatus(Address address, Status status)
+    {
+        PersistentState.SetStruct($"Status:{address}", status);
```
rowandh: Even though an enum is a struct, serializing them in this way will not work correctly. The recommendation for enums would be to cast to a `uint` and use `SetUInt32` to store them.

Testnet/Airdrop/Airdrop/Airdrop.cs
```diff
@@ -0,0 +1,202 @@
...
+
+    public struct StatusLog
+    {
+        [Index]
+        public Address Registrant;
+
+        public Status Status;
+    }
+
+    public enum Status
```
rowandh: Not necessary, but would recommend explicitly specifying the enum type here, `public enum Status : uint`

Testnet/Airdrop/Airdrop/Airdrop.cs
```diff
@@ -0,0 +1,202 @@
...
+
+        return true;
+    }
+
+    public struct StatusLog
+    {
+        [Index]
+        public Address Registrant;
+
+        public Status Status;
```
rowandh: Use a `uint` for Status here

Testnet/Airdrop/Airdrop/Airdrop.cs
```diff
@@ -0,0 +1,202 @@
...
+
+    /// <summary>Calculates and sets the amount to distribute to each registrant.</summary>
+    public ulong AmountToDistribute
+    {
+        get
+        {
+            ulong amount = PersistentState.GetUInt64(nameof(AmountToDistribute));
+            if (amount == 0 && RegistrationIsClosed)
+            {
+                amount = TotalSupply / NumberOfRegistrants;
```
rowandh: This will lose tokens due to rounding. Consider: `TotalSupply = 11` and `NumberOfRegistrants = 5`.

rowandh: To elaborate: the correct amount will still be distributed to the participants, but there will be some fraction of the total supply left over if the amounts are not evenly divisible. This might be OK if there are alternate means for the owner to withdraw those tokens from the contract.

Testnet/Airdrop/Airdrop/Airdrop.cs
```diff
@@ -0,0 +1,202 @@
...
+        get
+        {
+            bool isClosed = PersistentState.GetBool(nameof(RegistrationIsClosed));
+            if (!isClosed && EndBlock > 0)
+            {
+                isClosed = Block.Number > EndBlock;
+                RegistrationIsClosed = isClosed;
+            }
+
+            return isClosed;
```
rowandh: Would recommend moving this logic out of the getter. The main reason is that it makes the contract harder to reason about. There too many things are happening here - checking the block number, checking whether the registration is closed, and updating the `RegistrationIsClosed` property.

Consider breaking it into a few discrete steps:
- `RegistrationIsClosed` as a basic get/set property.
- Don't update `RegistrationIsClosed` when the block number is exceeded. Instead, compute this with a second property, something like `public bool CanRegister => RegistrationIsClosed || Block.Number > EndBlock;`.

Testnet/Airdrop/Airdrop/Airdrop.cs
```diff
@@ -0,0 +1,202 @@
...
+        }
+
+        private set => PersistentState.SetBool(nameof(RegistrationIsClosed), value);
+    }
+
+    /// <summary>Returns the status of any given address.</summary>
+    public Status GetAccountStatus(Address address)
```
rowandh: Return a `uint` here

Testnet/Airdrop/Airdrop/Airdrop.cs
```diff
@@ -0,0 +1,202 @@
...
+    {
+        bool invalidAccountStatus = GetAccountStatus(Message.Sender) != Status.ENROLLED;
+        if (invalidAccountStatus || !RegistrationIsClosed || AmountToDistribute == 0)
+        {
+            return false;
+        }
+
+        var transferParams = new object[] { Message.Sender, AmountToDistribute };
+
+        ITransferResult result = Call(TokenContractAddress, AmountToDistribute, "TransferTo", transferParams);
```
rowandh: Noting the earlier discussion about choice of `TransferTo` vs. `TransferFrom` - the intended usage of `TransferFrom` is exactly for scenarios like this, where a third party (or contract) needs to transfer tokens on another party's behalf. 

In short, `TransferTo` allows _you_ to transfer tokens that you own, while `TransferFrom` allows _someone else_ to transfer tokens that you own.

For a reason why this might be beneficial, consider the current scenario where `TransferTo` is used and some amount of tokens are assigned to the contract. For some reason, the airdrop is cancelled. Or, maybe there is a bug in the contract, and the tokens cannot be transferred to recipients. The Airdrop contract has no method to return those tokens to the owner, so they are effectively locked up and must either be distributed or burnt. If `TransferFrom` were used, the holder of the airdrop supply could just remove the approval for the old contract, and that would be it.

This affords the party running the airdrop a considerable amount of control, which could be a good or bad thing depending on the circumstances.

rowandh: Not suggesting one approach over the other. Both are valid, depending on the use case, but it's important to understand what happens if something does go wrong. Case in point: the rounding in `AmountToTransfer` could cause at least 1 token to be permanently burnt.

Testnet/Airdrop/Airdrop.Tests/AirdropTests.cs
```diff
@@ -0,0 +1,814 @@
...
+            // Owner returns owner supplied by param
+            MockPersistentState.Setup(x => x.GetAddress(nameof(Owner))).Returns(owner);
+            // TokenContractAddress returns tokenContractAddress supplied by param
+            MockPersistentState.Setup(x => x.GetAddress(nameof(TokenContractAddress))).Returns(TokenContractAddress);
+
+            // NumberOfRegistrants returns NumberOfRegistrants from this test class
+            MockPersistentState.Setup(x => x.GetUInt64(nameof(NumberOfRegistrants))).Returns(NumberOfRegistrants);
+            // NumberOfRegistrants, when set, increments the NumberOfRegistrations in this test class
+            MockPersistentState.Setup(x => x.SetUInt64(nameof(NumberOfRegistrants), It.IsAny<ulong>()))
+                .Callback<string, ulong>((key, value) => { NumberOfRegistrants += value; });
```
rowandh: Mocking the contract's implementation in this way will prevent any problems with the actual implementation being found. I'd recommend just setting `NumberOfRegistrants = value` here, and slightly modifying the test that the value is incremented.

rowandh:
Hey @mrtpain! The contract looks really good for the most part. The main change would be the fixes around serializing enums. We will work on improving our examples and docs around this.

I can see you've thought about TransferTo/From, and based on the tests the AmountToDistribute rounding behaviour seems to be expected, so feel free to ignore my comments on those. Do consider adding some additional docs around this behaviour for anyone who plans to use the contract.

mrtpain:
@rowandh @Neurosploit 

Hey guys, I've made what I believe are all the requested changes, refactoring, and updates to this contract. I'd appreciate one final review for approval when you have a moment.

I also added a readme which I think we should make a standard for all approvals. Maybe update the readme at the root of this repository to specify that. 

Any changes, let me know. If this isn't good to go, it should be very very close. Thanks!

Neurosploit:
Looks ok to me.

StratisIain:
Whitelisted on CirrusTest

PR #20, opened 2020-12-24
[Mainnet] Identity Contract

fenix2222:
# Description

Identity contract that enables decentralised claim management for users

# Compiler Version
`v.0.0.2`

# Hash
`60e8ac91c440d22daf04d65ff0cbd8ca173b5e78205fe939f01c7ef78970421b`

# Bytecode
```
4D5A90000300000004000000FFFF0000B800000000000000400000000000000000000000000000000000000000000000000000000000000000000000800000000E1FBA0E00B409CD21B8014CCD21546869732070726F6772616D2063616E6E6F742062652072756E20696E20444F53206D6F64652E0D0D0A2400000000000000504500004C010200018E6DB60000000000000000E00022200B013000000A00000002000000000000AE2900000020000000400000000000100020000000020000040000000000000004000000000000000060000000020000000000000300408500001000001000000000100000100000000000001000000000000000000000005C2900004F000000000000000000000000000000000000000000000000000000004000000C000000402900001C0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000200000080000000000000000000000082000004800000000000000000000002E74657874000000B409000000200000000A000000020000000000000000000000000000200000602E72656C6F6300000C0000000040000000020000000C000000000000000000000000000040000042000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000090290000000000004800000002000500D42100006C07000001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004602280500000A72010000706F0600000A2A4A02280500000A7201000070036F0700000A2A660203280800000A0202280900000A6F0A00000A28020000062A3A022805000006020328020000062A8A0202280100000602280900000A6F0A00000A280B00000A720D000070280C00000A2A00000013300400480000000100001102280500000602030428080000060A020304052809000006021201FE15040000021201037D040000041201047D050000041201057D060000041201067D0700000407280100002B2A13300300440000000200001102280500000602030428080000060A068E2D012A020304280A000006021201FE15030000021201037D010000041201047D020000041201067D0300000407280200002B2A8A02280500000A725D000070038C06000001048C0C000001280E00000A6F0F00000A2A8E02280500000A725D000070038C06000001048C0C000001280E00000A056F1000000A2A8A02280500000A725D000070038C06000001048C0C000001280E00000A6F1100000A2A000042534A4201000100000000000C00000076342E302E33303331390000000005006C000000F8020000237E0000640300009002000023537472696E677300000000F405000080000000235553007406000010000000234755494400000084060000E800000023426C6F6200000000000000020000015715A201090A000000FA013300160000010000000E00000004000000070000000A0000000F00000011000000070000000200000001000000010000000200000001000000020000000200000002000000000021010100000000000600C300E2010600F200E2010600AF00CF010F00020200000A00570239020A00310239020A00840039020600700033010A00E30039020A00980039020A005B003902060001003301060016013301060065023301000000000800000000000100010001001000840100001500010001000A0110004E000000210001000B000A01100041000000210004000B0006006C01810006003500850006001400880006006C0181000600350085000600140088000600110088005020000000008108A0012D0001006220000000008108AA018C0001007520000000008618C901220002008F20000000008600B4018C0003009E20000000008100730206000400C4200000000086003A0192000400182100000000860043019B00070068210000000086005A01A20009008B21000000008100630192000B00AF210000000081004F019B000E0000000100100100000100A90000000100C001000001007501000002003B00000003001900000001007501000002003B00000001007501000002003B00000001007501000002003B00000003001900000001007501000002003B000900C90101001100C90106001900C9010A004900C901060029007A00100051002302150051002E021B002900C9012200290064002800590095012D0031008302320029006C023A0029001D014700690050025F0051001102660051001A026C0051007E01730021002300E0002E000B00AF002E001300B8002E001B00D70041002300E00081002300E000A1002300E00040005300020001000000C301AA000200010003000100020003000480000000000000000000000000000000005702000004000000000000000000000078001E0000000000020000000000000000000000000039020000000003000200040002001B004E001B005A0000000055496E743332003C4D6F64756C653E004F6C644461746100646174610053797374656D2E507269766174652E436F72654C696200546F70696300746F70696300436C61696D4368616E67656400436C61696D52656D6F76656400494D657373616765006765745F4D6573736167650056616C756554797065006765745F53746174650049536D617274436F6E74726163745374617465004950657273697374656E7453746174650073746174650044656275676761626C6541747472696275746500436F6D70696C6174696F6E52656C61786174696F6E7341747472696275746500496E6465784174747269627574650052756E74696D65436F6D7061746962696C6974794174747269627574650076616C756500537472696E67004C6F6700536D617274436F6E74726163742E646C6C0053797374656D00416464436C61696D0052656D6F7665436C61696D00436C656172436C61696D00476574436C61696D00536574436C61696D00497373756564546F00697373756564546F00436C656172004964656E7469747950726F7669646572006765745F53656E646572006765745F4F776E6572007365745F4F776E6572004368616E67654F776E6572006E65774F776E6572002E63746F720053797374656D2E446961676E6F73746963730053797374656D2E52756E74696D652E436F6D70696C6572536572766963657300446562756767696E674D6F6465730047657442797465730053657442797465730047657441646472657373005365744164647265737300537472617469732E536D617274436F6E74726163747300466F726D617400536D617274436F6E7472616374004F626A6563740041737365727400456E737572654F776E65724F6E6C79006F705F457175616C6974790000000B4F0077006E0065007200004F54006800650020006D006500740068006F0064002000630061006E002000620065002000630061006C006C006500640020006200790020006F006E006C00790020006F0077006E00650072002E00001F43006C00610069006D005B007B0030007D005D005B007B0031007D005D0000000000DDF65F4D14FDB44894A2BB58E37AB9DB00042001010803200001052001011111042000122905200111190E062002010E111905200101121D042000122D0420001119070002021119111905200201020E0607021D05111006300101011E00040A0111100607021D05110C040A01110C0600030E0E1C1C0520011D050E062002010E1D05042001010E087CEC85D7BEA7798E0306111902060903061D05052001011119082003011119091D05062002011119090720021D0511190904280011190801000800000000001E01000100540216577261704E6F6E457863657074696F6E5468726F7773010801000200000000000401000000000000000000000000000000000000100000000000000000000000000000008429000000000000000000009E29000000200000000000000000000000000000000000000000000090290000000000000000000000005F436F72446C6C4D61696E006D73636F7265652E646C6C0000000000FF250020001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002000000C000000B03900000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000```

### CODE REVIEW

File: Mainnet/Identity/IdentityContracts/IdentityProvider.cs [hunk 1]
```diff
@@ -0,0 +1,143 @@
...
+        }
+        else
+        {
+            this.Log(new ClaimAdded
+            {
+                IssuedTo = issuedTo,
+                Topic = topic,
+                Data = data
+            });
+        }
```
mrtpain, line 62: Nit pick, also depends on current internal use cases that may already exist but...

These logs for `ClaimChanged, ClaimAdded, and ClaimRemoved` all share the same shape but since they are different, querying this contracts receipts would require a separate Cirrus Local Call for each event to check the logs of all 3 types. Not breaking but more difficult to use in my experience.

**Suggestion**
Use 1 new log type of `ClaimEvent`, adding a new property for `Type` and Index that in the event.

``` C#
// New single event type
public struct ClaimEvent
{
    [Index]
    public Address IssuedTo;
    [Index]
    public uint Topic;
    public byte[] Data;
    [Index]
    public ClaimType Type;
}

// Type enum 
public enum ClaimType : int
{
    Added = 0,
    Changed = 1,
    Removed = 2
}

// Log accordingly
this.Log(new ClaimAdded
{
    IssuedTo = issuedTo,
    Topic = topic,
    Data = data,
    Type = ClaimType.Added // or any other claim type
});
```

mrtpain (commented):
Overall, bytecode and hash matches when compiled locally. Tests pass and code in general looks ok.

I have one suggestion around the Claims logs but is not a blocker, just a general improvement I think would be good to adjust for. Thoughts?

mrtpain (approved):
I think combining the logged events to one object since they are so similar would be a big benefit when querying the contract, but is not critical. Up to you, approving as is

Mainnet/Identity/IdentityContracts/IdentityProvider.cs
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    public void RemoveClaim(Address issuedTo, uint topic)
+    {
+        Assert(this.Owner == Message.Sender);
+
+        // Nothing to delete.
+        byte[] oldData = this.GetClaim(issuedTo, topic);
+        if (oldData == null)
```
rowandh, line 71: Value can never be null here because `GetBytes` will return `byte[0]` for empty values. Should be fine to remove, or check `oldData.Length == 0`

Mainnet/Identity/IdentityContracts/IdentityProvider.cs
```diff
@@ -0,0 +1,143 @@
...
+        Assert(this.Owner == Message.Sender);
+        this.Owner = newOwner;
+    }
+
+    public void AddClaim(Address issuedTo, uint topic, byte[] data)
+    {
+        Assert(this.Owner == Message.Sender);
+
+        byte[] oldData = this.GetClaim(issuedTo, topic);
+        bool exists = oldData != null;
```
rowandh, line 37: Same as the below comment - `exists` will always be true here.

Mainnet/Identity/IdentityContracts/IdentityProvider.cs
```diff
@@ -0,0 +1,143 @@
...
+
+        // This exact claim value is stored already. Nothing to change.
+        if (exists && ByteArrayCompare(oldData, data))
```
rowandh, line 40: The byte array comparison allows us to know if the claim has been changed. Is that valuable, or could we just log `ClaimChanged` any time the method is invoked, even if the claim is the same?

The downsides I see are added complexity by requiring the byte array comparison, which is unbounded and could potentially consume `GasLimitMaximum` if the arrays are big enough.

Mainnet/Identity/IdentityContracts/IdentityProvider.cs
```diff
@@ -0,0 +1,143 @@
...
+
+    public void ChangeOwner(Address newOwner)
+    {
+        Assert(this.Owner == Message.Sender);
+        this.Owner = newOwner;
+    }
+
+    public void AddClaim(Address issuedTo, uint topic, byte[] data)
+    {
+        Assert(this.Owner == Message.Sender);
```
rowandh, line 34: Enforcing a maximum length on `data` here would be a good idea.

rowandh:
A few issues which don't affect the overall functionality but could be gotchas in the future.

Mainnet/Identity/IdentityContracts/IdentityProvider.cs
YakupIpek: On receipt/search endpoint, we specify log filter by name so it is important to differentiate log names.

Mainnet/Identity/IdentityContracts/IdentityProvider.cs
YakupIpek: We can leave this responsibility to client and save some gas so client can compare it with previous data if it really wants to know it is new data or not.

Mainnet/Identity/IdentityContracts/IdentityProvider.cs
YakupIpek: What would be max length ?

YakupIpek:
@rowandh I pushed my changes. It looks fine ?

Mainnet/Identity/IdentityContracts/IdentityProvider.cs
rowandh: Up to you to decide. I would push this to the client myself and log `OldData` so the client can perform the comparison :)

Mainnet/Identity/IdentityContracts/IdentityProvider.cs
mrtpain: You can have 1 `ClaimEvent` that returns all claim types and just filter the JSON returned for the ones you wanted. otherwise if you ever wanted all types, you'll have to query for each type individually. Up to you guys, sometimes easier to combine like shaped objects.

rowandh:
Compiled locally:
Hash: `60e8ac91c440d22daf04d65ff0cbd8ca173b5e78205fe939f01c7ef78970421b`

Bytecode:
```
4D5A90000300000004000000FFFF0000B800000000000000400000000000000000000000000000000000000000000000000000000000000000000000800000000E1FBA0E00B409CD21B8014CCD21546869732070726F6772616D2063616E6E6F742062652072756E20696E20444F53206D6F64652E0D0D0A2400000000000000504500004C010200018E6DB60000000000000000E00022200B013000000A00000002000000000000AE2900000020000000400000000000100020000000020000040000000000000004000000000000000060000000020000000000000300408500001000001000000000100000100000000000001000000000000000000000005C2900004F000000000000000000000000000000000000000000000000000000004000000C000000402900001C0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000200000080000000000000000000000082000004800000000000000000000002E74657874000000B409000000200000000A000000020000000000000000000000000000200000602E72656C6F6300000C0000000040000000020000000C000000000000000000000000000040000042000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000090290000000000004800000002000500D42100006C07000001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004602280500000A72010000706F0600000A2A4A02280500000A7201000070036F0700000A2A660203280800000A0202280900000A6F0A00000A28020000062A3A022805000006020328020000062A8A0202280100000602280900000A6F0A00000A280B00000A720D000070280C00000A2A00000013300400480000000100001102280500000602030428080000060A020304052809000006021201FE15040000021201037D040000041201047D050000041201057D060000041201067D0700000407280100002B2A13300300440000000200001102280500000602030428080000060A068E2D012A020304280A000006021201FE15030000021201037D010000041201047D020000041201067D0300000407280200002B2A8A02280500000A725D000070038C06000001048C0C000001280E00000A6F0F00000A2A8E02280500000A725D000070038C06000001048C0C000001280E00000A056F1000000A2A8A02280500000A725D000070038C06000001048C0C000001280E00000A6F1100000A2A000042534A4201000100000000000C00000076342E302E33303331390000000005006C000000F8020000237E0000640300009002000023537472696E677300000000F405000080000000235553007406000010000000234755494400000084060000E800000023426C6F6200000000000000020000015715A201090A000000FA013300160000010000000E00000004000000070000000A0000000F00000011000000070000000200000001000000010000000200000001000000020000000200000002000000000021010100000000000600C300E2010600F200E2010600AF00CF010F00020200000A00570239020A00310239020A00840039020600700033010A00E30039020A00980039020A005B003902060001003301060016013301060065023301000000000800000000000100010001001000840100001500010001000A0110004E000000210001000B000A01100041000000210004000B0006006C01810006003500850006001400880006006C0181000600350085000600140088000600110088005020000000008108A0012D0001006220000000008108AA018C0001007520000000008618C901220002008F20000000008600B4018C0003009E20000000008100730206000400C4200000000086003A0192000400182100000000860043019B00070068210000000086005A01A20009008B21000000008100630192000B00AF210000000081004F019B000E0000000100100100000100A90000000100C001000001007501000002003B00000003001900000001007501000002003B00000001007501000002003B00000001007501000002003B00000003001900000001007501000002003B000900C90101001100C90106001900C9010A004900C901060029007A00100051002302150051002E021B002900C9012200290064002800590095012D0031008302320029006C023A0029001D014700690050025F0051001102660051001A026C0051007E01730021002300E0002E000B00AF002E001300B8002E001B00D70041002300E00081002300E000A1002300E00040005300020001000000C301AA000200010003000100020003000480000000000000000000000000000000005702000004000000000000000000000078001E0000000000020000000000000000000000000039020000000003000200040002001B004E001B005A0000000055496E743332003C4D6F64756C653E004F6C644461746100646174610053797374656D2E507269766174652E436F72654C696200546F70696300746F70696300436C61696D4368616E67656400436C61696D52656D6F76656400494D657373616765006765745F4D6573736167650056616C756554797065006765745F53746174650049536D617274436F6E74726163745374617465004950657273697374656E7453746174650073746174650044656275676761626C6541747472696275746500436F6D70696C6174696F6E52656C61786174696F6E7341747472696275746500496E6465784174747269627574650052756E74696D65436F6D7061746962696C6974794174747269627574650076616C756500537472696E67004C6F6700536D617274436F6E74726163742E646C6C0053797374656D00416464436C61696D0052656D6F7665436C61696D00436C656172436C61696D00476574436C61696D00536574436C61696D00497373756564546F00697373756564546F00436C656172004964656E7469747950726F7669646572006765745F53656E646572006765745F4F776E6572007365745F4F776E6572004368616E67654F776E6572006E65774F776E6572002E63746F720053797374656D2E446961676E6F73746963730053797374656D2E52756E74696D652E436F6D70696C6572536572766963657300446562756767696E674D6F6465730047657442797465730053657442797465730047657441646472657373005365744164647265737300537472617469732E536D617274436F6E74726163747300466F726D617400536D617274436F6E7472616374004F626A6563740041737365727400456E737572654F776E65724F6E6C79006F705F457175616C6974790000000B4F0077006E0065007200004F54006800650020006D006500740068006F0064002000630061006E002000620065002000630061006C006C006500640020006200790020006F006E006C00790020006F0077006E00650072002E00001F43006C00610069006D005B007B0030007D005D005B007B0031007D005D0000000000DDF65F4D14FDB44894A2BB58E37AB9DB00042001010803200001052001011111042000122905200111190E062002010E111905200101121D042000122D0420001119070002021119111905200201020E0607021D05111006300101011E00040A0111100607021D05110C040A01110C0600030E0E1C1C0520011D050E062002010E1D05042001010E087CEC85D7BEA7798E0306111902060903061D05052001011119082003011119091D05062002011119090720021D0511190904280011190801000800000000001E01000100540216577261704E6F6E457863657074696F6E5468726F7773010801000200000000000401000000000000000000000000000000000000100000000000000000000000000000008429000000000000000000009E29000000200000000000000000000000000000000000000000000090290000000000000000000000005F436F72446C6C4D61696E006D73636F7265652E646C6C0000000000FF250020001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002000000C000000B03900000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
```

@YakupIpek could you please verify these and update the readme, then we can merge.

YakupIpek:
Pr is not belong to me so I can't update description

YakupIpek:
Your hash and hash in readme files matches. I am not able to update pr description

Mainnet/Identity/IdentityContracts/IdentityProvider.cs
```diff
@@ -0,0 +1,106 @@
...
+
+    public void ChangeOwner(Address newOwner)
+    {
+        EnsureOwnerOnly();
+        Owner = newOwner;
+    }
+
+    private void EnsureOwnerOnly()
+    {
+        Assert(Owner == Message.Sender,"The method can be called by only owner.");
```
dangershony, line 28: This message seems wrong
"The method can be called only by owner"

Mainnet/Identity/IdentityContracts/IdentityProvider.cs
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    public void AddClaim(Address issuedTo, uint topic, byte[] data)
```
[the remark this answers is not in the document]
YakupIpek: I don't think that limitation is needed. User can push data as long as gas limit is not reached.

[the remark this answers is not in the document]
dangershony: Ah the limitation is on the fee fair enough, however what is supp0ose to go in this parameter?
From history we learn that sometimes the end apps like wallets assume a limitation and may break

PR #25, opened 2020-12-31
[Mainnet] Address mapper

YakupIpek:
# Address Mapper Contract

**Compiler Version**
```
v2.0.0
```
**Contract Hash**
```
ea55b6d2485904e940c1495472b3d25d1b2c01161f68714de02da8278f2635f1
```

**Contract Byte Code**
```
4D5A90000300000004000000FFFF0000B800000000000000400000000000000000000000000000000000000000000000000000000000000000000000800000000E1FBA0E00B409CD21B8014CCD21546869732070726F6772616D2063616E6E6F742062652072756E20696E20444F53206D6F64652E0D0D0A2400000000000000504500004C01020042542C900000000000000000E00022200B013000000E00000002000000000000C22C0000002000000040000000000010002000000002000004000000000000000400000000000000006000000002000000000000030040850000100000100000000010000010000000000000100000000000000000000000702C00004F000000000000000000000000000000000000000000000000000000004000000C000000542C00001C0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000200000080000000000000000000000082000004800000000000000000000002E74657874000000C80C000000200000000E000000020000000000000000000000000000200000602E72656C6F6300000C000000004000000002000000100000000000000000000000000000400000420000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000A42C0000000000004800000002000500B0220000A409000001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000007202280500000A7201000070038C06000001280600000A6F0700000A2A7602280500000A7201000070038C06000001280600000A046F0800000A2A7202280500000A722B000070038C06000001280600000A6F0100002B2A7602280500000A722B000070038C06000001280600000A046F0200002B2A7202280500000A7243000070038C06000001280600000A6F0B00000A2A4602280500000A72630000706F0700000A2A4A02280500000A7263000070036F0800000A2A3E0203280C00000A020428070000062A13300400450000000100001102020328030000067B0600000416FE01726F000070280D00000A02031200FE1504000002120002280E00000A6F0F00000A7D050000041200177D060000040628040000062A00000013300400790000000200001102280E000006020328030000060A02067B0600000417FE01728D000070280D00000A02067B0500000403280200000602031201FE15040000021201067B050000047D050000041201187D06000004072804000006021202FE15050000021202067B050000047D070000041202037D0800000408280300002B2A000000133003002A0000000100001102280E000006020328030000060A02067B0600000417FE01728D000070280D00000A020328050000062A0000133003002300000001000011020328030000060A02067B0600000418FE0172CF000070280D00000A067B050000042ABA02280E00000602037E1100000A281200000A72090100707271010070280600000A280D00000A020328070000062A8A0202280600000602280E00000A6F0F00000A281300000A727D010070280D00000A2A00000042534A4201000100000000000C00000076342E302E33303331390000000005006C00000068030000237E0000D40300000003000023537472696E677300000000D4060000C00100002355530094080000100000002347554944000000A40800000001000023426C6F620000000000000002000001571DA201090A000000FA013300160000010000000E00000005000000080000000E0000000F0000001300000003000000060000000200000001000000010000000200000001000000020000000300000003000000000038010100000000000600A700CA010600D600CA0106009300B7010F00EA0100000A007B0254020A004C0254020A005B005402060051014A01060047004A010A00C70054020A0082005402060020014A01060089024A010A0032005402000000000100000000000100010001001000A3010000150001000100020100006D020000210001000F000A0110005B010000250005000F000A01100027010000250007000F0006060A00950056806B0298005680020198005680290098000600D602710006006D0295000600D60271000600C202710050200000000086001A029C0001006D200000000081002E02A30002008B200000000086000A01AB000400A8200000000081001501B2000500C6200000000081005601BA000700E3200000000086087D0157000800F5200000000081088701BA0008000821000000008618B101C00009001821000000008600F901BA000B006C21000000008600FA00BA000C00F4210000000086009002BA000D002C2200000000860042029C000E005B220000000086009101BA000F008A22000000008600B2020600100000000100DE0200000100DE0200000200CC0200000100CC0200000100CC0200000200F40000000100CC0200000100F400000001006F00000002009D0100000100CC0200000100CC0200000100CC0200000100CC02000001009D010900B10101001100B10106001900B1010A005100B1010600290051001000610074021500590004021B0059000F0221005900970228005900A102340059006C013C002900B10141002900AB024C0029003B0052007100720157002900340165003100670171003100F20275003100E602750008000800860008000C008B000800100090002E000B00CD002E001300D6002E001B00F500A10023008B00E10023008B00010123008B0047005C000200010000009701C8000200060003000100070003000480000000000000000000000000000000007B0200000400000000000000000000007D00120000000000020000000000000000000000000054020000000003000200040002000500020013002F0015002F0021006C000000003C4D6F64756C653E0076616C75655F5F0053797374656D2E507269766174652E436F72654C696200417070726F76656400494D657373616765006765745F4D6573736167650056616C756554797065006765745F53746174650049536D617274436F6E7472616374537461746500736D617274436F6E74726163745374617465004950657273697374656E7453746174650044656275676761626C6541747472696275746500436F6D70696C6174696F6E52656C61786174696F6E7341747472696275746500496E6465784174747269627574650052756E74696D65436F6D7061746962696C6974794174747269627574650076616C756500417070726F76650050656E64696E67004765744D617070696E67005365744D617070696E6700537472696E6700416464726573734D61707065644C6F6700536D617274436F6E74726163742E646C6C0053797374656D00456E756D00436C6561724D617070696E67496E666F005A65726F00436C656172006765745F53656E646572006765745F4F776E6572007365745F4F776E6572004368616E67654F776E6572006F776E657200416464726573734D6170706572002E63746F720053797374656D2E446961676E6F73746963730053797374656D2E52756E74696D652E436F6D70696C6572536572766963657300446562756767696E674D6F646573004D61704164647265737300476574416464726573730053657441646472657373004765745365636F6E6461727941646472657373005365745365636F6E6461727941646472657373004765745072696D6172794164647265737300537472617469732E536D617274436F6E747261637473004E6F53746174757300466F726D617400536D617274436F6E7472616374004F626A6563740052656A65637400476574537472756374005365745374727563740041737365727400456E7375726541646D696E4F6E6C79005365636F6E64617279007365636F6E64617279005072696D617279007072696D617279006F705F457175616C697479006F705F496E657175616C6974790000295300650063006F006E00640061007200790041006400640072006500730073003A007B0030007D0000174D0061007000700069006E0067003A007B0030007D00001F4D0061007000700069006E00670049006E0066006F003A007B0030007D00000B4F0077006E0065007200001D41007300730065007200740020006600610069006C00650064002E0000414D0061007000700069006E00670020006900730020006E006F007400200069006E002000700065006E00640069006E0067002000730074006100740065002E00003954006800650020006D0061007000700069006E00670020006900730020006E006F007400200061007000700072006F007600650064002E00006754006800650020007B0030007D00200070006100720061006D0065007400650072002000630061006E0020006E006F0074002000620065002000640065006600610075006C00740028007A00650072006F002900200061006400640072006500730073002E00000B6F0077006E0065007200003F4F006E006C007900200063006F006E007400720061006300740020006F0077006E00650072002000630061006E0020006100630063006500730073002E0000000000EEE2C7F974D6C54DB86E6FFA481534DB00042001010803200001052001011111042000122D0500020E0E1C05200111190E062002010E1119063001011E000E040A01111007300102010E1E00042001010E05200101121D040701111005200201020E0420001239042000111908070311101110111406300101011E00040A011114030611190700020211191119087CEC85D7BEA7798E0400000000040100000004020000000206080306110C06200111191119072002011119111906200111101119072002011119111005200101111907200201121D111904280011190801000800000000001E01000100540216577261704E6F6E457863657074696F6E5468726F777301080100020000000000000000000000000000000000000010000000000000000000000000000000982C00000000000000000000B22C0000002000000000000000000000000000000000000000000000A42C0000000000000000000000005F436F72446C6C4D61696E006D73636F7265652E646C6C0000000000FF2500200010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002000000C000000C43C00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
```


mrtpain:
Should Testnet files be included in this PR? There is a separate PR for Testnet Address Mapper https://github.com/stratisproject/CirrusSmartContracts/pull/18. Not sure which to follow.

Mainnet/AddressMapper/AddressMapper/AddressMapper.cs
```diff
@@ -0,0 +1,78 @@
...
+
+    public Address Owner
+    {
+        get => PersistentState.GetAddress(nameof(Owner));
+        private set => PersistentState.SetAddress(nameof(Owner), value);
+    }
+
+    public AddressMapper(ISmartContractState smartContractState, Address Owner) : base(smartContractState)
+    {
+        this.Owner = Owner;
```
mrtpain, line 22: Just verifying, the intention here is to allow any owner by sending an address on contract creation and not make the owner `Message.Sender` by default?

Mainnet/AddressMapper/AddressMapper/AddressMapper.cs
```diff
@@ -0,0 +1,78 @@
...
+    private MappingInfo GetMappingInfo(Address secondary) => PersistentState.GetStruct<MappingInfo>($"MappingInfo:{secondary}");
+    private void SetMappingInfo(Address secondary, MappingInfo value) => PersistentState.SetStruct($"MappingInfo:{secondary}", value);
+    
+    private void ClearMappingInfo(Address secondary) => PersistentState.Clear($"MappingInfo:{secondary}");
+
+    public Address Owner
+    {
+        get => PersistentState.GetAddress(nameof(Owner));
+        private set => PersistentState.SetAddress(nameof(Owner), value);
+    }
```
mrtpain, line 18: There probably should be a public method that allows the "Owner" to set a new `Owner` value, or possibly multiple `Approver` values. 

I see `MapAddress` allows anyone to map a secondary address to the address then sent the Transaction from, which then requires `Approve` or `Reject` by the `Owner`. One wallet or owner to handle them all might cause a backlog, all funneled through 1 person.

Mainnet/AddressMapper/AddressMapper.Tests/AddressMapperTests.cs
```diff
@@ -0,0 +1,145 @@
...
+            contractAddress = "0x0000000000000000000000000000000000000001".HexToAddress();
+            ownerAddress = "0x0000000000000000000000000000000000000002".HexToAddress();
+            primaryAddress = "0x0000000000000000000000000000000000000003".HexToAddress();
+            secondaryAddress = "0x0000000000000000000000000000000000000004".HexToAddress();
+        }
+
+        [Theory]
+        [InlineData(Status.Approved)]
+        [InlineData(Status.Pending)]
+        public void MappAddress_Fails_If_SecondaryAddress_Is_Mapped_Already(Status status)
```
mrtpain, line 40: I don't know the actual use case for these mappings but _should_ there be a way for this to be changed?

Mainnet/AddressMapper/AddressMapper.Tests/AddressMapperTests.cs
```diff
@@ -0,0 +1,145 @@
...
+        {
+            persistentState.SetStruct($"MappingInfo:{secondaryAddress}", new MappingInfo { Primary = primaryAddress, Status = (int)Status.NoStatus });
+
+            mContractState.Setup(x => x.Message).Returns(new Message(contractAddress, primaryAddress, 0));
+
+            var contract = new AddressMapper(mContractState.Object, ownerAddress);
+
+            contract.MapAddress(secondaryAddress);
+
+            Assert.Equal(new MappingInfo { Primary = primaryAddress, Status = (int)Status.Pending }, persistentState.GetStruct<MappingInfo>($"MappingInfo:{secondaryAddress}"));
```
mrtpain, line 62: Because you set the struct on Line 54 above, line 62 is always going to be true. 

It would be better to use 
```
mContractState.Verify(x => x.SetStruct<MappingInfo>(...), Times.Once);
```

to validate that the contract call actually made it there in code and persisted the correct values on line 61 after you call the contract method.

Mainnet/AddressMapper/AddressMapper.Tests/AddressMapperTests.cs
```diff
@@ -0,0 +1,145 @@
...
+
+            var contract = new AddressMapper(mContractState.Object, ownerAddress);
+
+            contract.Approve(secondaryAddress);
+
+            Assert.Equal(secondaryAddress, persistentState.GetAddress($"SecondaryAddress:{primaryAddress}"));
+
+            Assert.Equal(new MappingInfo { Primary = primaryAddress, Status = (int)Status.Approved }, persistentState.GetStruct<MappingInfo>($"MappingInfo:{secondaryAddress}"));
+
+            mContractLogger.Verify(m => m.Log(mContractState.Object, new AddressMappedLog { Primary = primaryAddress, Secondary = secondaryAddress }));
```
mrtpain, line 104: Same concept here, probably better to assert and verify that persistent state was actually set rather than the checking these values equal each other.

mrtpain:
Overall, I have some questions on some of the approaches in the contract as well as how some of the tests are written and asserted. LGTM for the most part but think some things could be improved to harden the contract and allow for some edge case scenarios and use cases.

YakupIpek:
> Should Testnet files be included in this PR? There is a separate PR for Testnet Address Mapper #18. Not sure which to follow.

It is actually not good idea to push Testnet and Mainnet pull requests together. It is confusing. More proper way is to review and accept Testnet firstly then create Mainnet. Otherwise pr will contains both files. Alternatively we can seperate branches but this time it is weird to accept Mainnet pr first. 

It is ok for now if you can review only for Mainnet files.

Mainnet/AddressMapper/AddressMapper/AddressMapper.cs
YakupIpek: `Approve` and `Reject `methods are called by address owners from Identity app. It is responsibility of the Identity app to ask user permission when a mapping requested to user's address. Identity server will handle calling Approve and Reject methods. This is not a manual process for the contract owner.  This is not going to cause backlog there. 

I added ChangeOwner method for set new Owner.

Mainnet/AddressMapper/AddressMapper.Tests/AddressMapperTests.cs
YakupIpek: Line 54 sets `NoStatus` for `MappingInfo` and line 62 is verify that it is set to `Pending` state after `MappAddress`called. 

I just realized that line 54 is unnecessary. I removed it. 

I can not call verify method because persistentState is not a mocked object. It is InMemoryState for easier testing.

Mainnet/AddressMapper/AddressMapper.Tests/AddressMapperTests.cs
YakupIpek: `persistentState` is not a mocked object. It is `InMemoryState`.

mrtpain (approved):
Nice work, this looks good to me 👍

mrtpain (changes requested):
Actually one last change, the hash and bytecode just need to be updated, then I can verify locally and approve

Mainnet/AddressMapper/AddressMapper/AddressMapper.cs
YakupIpek: Yes it is intentionally designed this way

Mainnet/AddressMapper/AddressMapper.Tests/AddressMapperTests.cs
YakupIpek: How ?

mrtpain (approved):
I see mainnet readme update just need to update the testnet contract readme bytecode / hash and the PR details bytecode / hash 👍

Mainnet/AddressMapper/AddressMapper/AddressMapper.cs
```diff
@@ -0,0 +1,88 @@
...
+
+    public AddressMapper(ISmartContractState smartContractState, Address owner) : base(smartContractState)
+    {
+        this.Owner = owner;
+    }
+
+    public void MapAddress(Address secondary)
+    {
```
rowandh, line 26: Should it be possible to map an address to itself? Ie. if `secondary == Message.Sender`?

Mainnet/AddressMapper/AddressMapper/AddressMapper.cs
```diff
@@ -0,0 +1,88 @@
...
+        EnsureAdminOnly();
+        var mapping = GetMappingInfo(secondary);
+        Assert(mapping.Status == (int)Status.Pending, "Mapping is not in pending state.");
+
+        ClearMappingInfo(secondary); // same address can be mapped again. 
+    }
+
+    public string GetStatus(Address secondary) => GetMappingInfo(secondary).Status.ToString();
+    public Address GetPrimaryAddress(Address secondary) => GetMappingInfo(secondary).Primary;
```
rowandh, line 54: This will map the secondary=>primary address even if the mapping has not been approved. This is inconsistent: we can't map primary=>secondary by calling `GetSecondaryAddress` until the mapping is approved.

Mainnet/AddressMapper/AddressMapper/AddressMapper.cs
rowandh, line 26: Anyone can DoS this contract by attempting to register a secondary address that doesn't belong to them. Although it requires approval, it will prevent the valid owner from being able to make a claim. Theoretically a malicious actor who really wants to prevent someone from registering an address could watch for a rejection tx in a block and repeat the process indefinitely.

rowandh:
I think `GetPrimaryAddress` needs to check for a valid approval.

The DoS vector may not be likely in practice but would be easy to accomplish. And it's avoidable with a design that allows for multiple mappings for a secondary address, leaving it up to the owner to decide which is real and which is not.

rowandh:
One more point (this may be by design): mappings are permanent because there's no method for the owner to release or update them.

Mainnet/AddressMapper/AddressMapper/AddressMapper.cs
YakupIpek: Identity Server is owner of the contract and Approve/Reject mapping requests based on logged in user decisions. So basically user approve/reject the mapping over Identity Server. 

User wants to associate his desktop address to mobile address so user has to make a call to contract from Desktop wallet by passing mobile wallet address. So for this reason method is not owner only. 

Secondary address is mobile address and it is very hard to know who is the address owner also it is easy to generate new account on mobile app. Theoretically possible but practically very hard abuse the contract. Also DOS attack is going to cost to attacker.

YakupIpek: I will look is there a better solution.

Mainnet/AddressMapper/AddressMapper/AddressMapper.cs
YakupIpek: > I think `GetPrimaryAddress` needs to check for a valid approval.
> 
> The DoS vector may not be likely in practice but would be easy to accomplish. And it's avoidable with a design that allows for multiple mappings for a secondary address, leaving it up to the owner to decide which is real and which is not.

Currently `GetPrimaryAddress` return the result whether it is approved or pending. So client can get status by calling `GetStatus`. This is how Identity and STO Platform works. But method naming is confusing bit of. I think this can be design as you requested way as well. 

There was reason we have gone in this style of design but i think this is not necessary after seeing use case of both clients so i will update the code as you requested way.

YakupIpek:
> One more point (this may be by design): mappings are permanent because there's no method for the owner to release or update them.

Yes currently we stick to immutable design for approved mappings. But it can be designed other way as well if there is a requirement.

Mainnet/AddressMapper/AddressMapper/AddressMapper.cs
YakupIpek: Well we haven't restricted to code for that case but it is not going to harm the contract. We add this assertion so will add extra gas for users who sent parameter correctly. 

Do you prefer to add assertion for this ?

Mainnet/AddressMapper/AddressMapper/AddressMapper.csproj
```diff
@@ -0,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+	
+	<LangVersion>8.0</LangVersion>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Stratis.SmartContracts" Version="2.0.1" />
```
rowandh, line 9: I believe this should be `2.0.0` as `2.0.1` is a dev release only

rowandh:
@YakupIpek could you please remove or make the testnet contract the same as mainnet?

mainnet/addressmapper/AddressMapper/AddressMapper.cs
```diff
@@ -0,0 +1,95 @@
...
+    public void Reject(Address secondary)
+    {
+        EnsureAdminOnly();
+        var mapping = GetMappingInfo(secondary);
+        Assert(mapping.Status == (int)Status.Pending, "Mapping is not in pending state.");
+
+        ClearMappingInfo(secondary); // same address can be mapped again. 
+    }
+
+    public string GetStatus(Address secondary) => GetMappingInfo(secondary).Status.ToString();
```
rowandh, line 53: Just checking this is desired - this will return "0" and not "NoStatus"

YakupIpek: Fixed

rowandh:
Compiled locally with these results:

Hash: `ea55b6d2485904e940c1495472b3d25d1b2c01161f68714de02da8278f2635f1`

Bytecode: 
```
4D5A90000300000004000000FFFF0000B800000000000000400000000000000000000000000000000000000000000000000000000000000000000000800000000E1FBA0E00B409CD21B8014CCD21546869732070726F6772616D2063616E6E6F742062652072756E20696E20444F53206D6F64652E0D0D0A2400000000000000504500004C01020042542C900000000000000000E00022200B013000000E00000002000000000000C22C0000002000000040000000000010002000000002000004000000000000000400000000000000006000000002000000000000030040850000100000100000000010000010000000000000100000000000000000000000702C00004F000000000000000000000000000000000000000000000000000000004000000C000000542C00001C0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000200000080000000000000000000000082000004800000000000000000000002E74657874000000C80C000000200000000E000000020000000000000000000000000000200000602E72656C6F6300000C000000004000000002000000100000000000000000000000000000400000420000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000A42C0000000000004800000002000500B0220000A409000001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000007202280500000A7201000070038C06000001280600000A6F0700000A2A7602280500000A7201000070038C06000001280600000A046F0800000A2A7202280500000A722B000070038C06000001280600000A6F0100002B2A7602280500000A722B000070038C06000001280600000A046F0200002B2A7202280500000A7243000070038C06000001280600000A6F0B00000A2A4602280500000A72630000706F0700000A2A4A02280500000A7263000070036F0800000A2A3E0203280C00000A020428070000062A13300400450000000100001102020328030000067B0600000416FE01726F000070280D00000A02031200FE1504000002120002280E00000A6F0F00000A7D050000041200177D060000040628040000062A00000013300400790000000200001102280E000006020328030000060A02067B0600000417FE01728D000070280D00000A02067B0500000403280200000602031201FE15040000021201067B050000047D050000041201187D06000004072804000006021202FE15050000021202067B050000047D070000041202037D0800000408280300002B2A000000133003002A0000000100001102280E000006020328030000060A02067B0600000417FE01728D000070280D00000A020328050000062A0000133003002300000001000011020328030000060A02067B0600000418FE0172CF000070280D00000A067B050000042ABA02280E00000602037E1100000A281200000A72090100707271010070280600000A280D00000A020328070000062A8A0202280600000602280E00000A6F0F00000A281300000A727D010070280D00000A2A00000042534A4201000100000000000C00000076342E302E33303331390000000005006C00000068030000237E0000D40300000003000023537472696E677300000000D4060000C00100002355530094080000100000002347554944000000A40800000001000023426C6F620000000000000002000001571DA201090A000000FA013300160000010000000E00000005000000080000000E0000000F0000001300000003000000060000000200000001000000010000000200000001000000020000000300000003000000000038010100000000000600A700CA010600D600CA0106009300B7010F00EA0100000A007B0254020A004C0254020A005B005402060051014A01060047004A010A00C70054020A0082005402060020014A01060089024A010A0032005402000000000100000000000100010001001000A3010000150001000100020100006D020000210001000F000A0110005B010000250005000F000A01100027010000250007000F0006060A00950056806B0298005680020198005680290098000600D602710006006D0295000600D60271000600C202710050200000000086001A029C0001006D200000000081002E02A30002008B200000000086000A01AB000400A8200000000081001501B2000500C6200000000081005601BA000700E3200000000086087D0157000800F5200000000081088701BA0008000821000000008618B101C00009001821000000008600F901BA000B006C21000000008600FA00BA000C00F4210000000086009002BA000D002C2200000000860042029C000E005B220000000086009101BA000F008A22000000008600B2020600100000000100DE0200000100DE0200000200CC0200000100CC0200000100CC0200000200F40000000100CC0200000100F400000001006F00000002009D0100000100CC0200000100CC0200000100CC0200000100CC02000001009D010900B10101001100B10106001900B1010A005100B1010600290051001000610074021500590004021B0059000F0221005900970228005900A102340059006C013C002900B10141002900AB024C0029003B0052007100720157002900340165003100670171003100F20275003100E602750008000800860008000C008B000800100090002E000B00CD002E001300D6002E001B00F500A10023008B00E10023008B00010123008B0047005C000200010000009701C8000200060003000100070003000480000000000000000000000000000000007B0200000400000000000000000000007D00120000000000020000000000000000000000000054020000000003000200040002000500020013002F0015002F0021006C000000003C4D6F64756C653E0076616C75655F5F0053797374656D2E507269766174652E436F72654C696200417070726F76656400494D657373616765006765745F4D6573736167650056616C756554797065006765745F53746174650049536D617274436F6E7472616374537461746500736D617274436F6E74726163745374617465004950657273697374656E7453746174650044656275676761626C6541747472696275746500436F6D70696C6174696F6E52656C61786174696F6E7341747472696275746500496E6465784174747269627574650052756E74696D65436F6D7061746962696C6974794174747269627574650076616C756500417070726F76650050656E64696E67004765744D617070696E67005365744D617070696E6700537472696E6700416464726573734D61707065644C6F6700536D617274436F6E74726163742E646C6C0053797374656D00456E756D00436C6561724D617070696E67496E666F005A65726F00436C656172006765745F53656E646572006765745F4F776E6572007365745F4F776E6572004368616E67654F776E6572006F776E657200416464726573734D6170706572002E63746F720053797374656D2E446961676E6F73746963730053797374656D2E52756E74696D652E436F6D70696C6572536572766963657300446562756767696E674D6F646573004D61704164647265737300476574416464726573730053657441646472657373004765745365636F6E6461727941646472657373005365745365636F6E6461727941646472657373004765745072696D6172794164647265737300537472617469732E536D617274436F6E747261637473004E6F53746174757300466F726D617400536D617274436F6E7472616374004F626A6563740052656A65637400476574537472756374005365745374727563740041737365727400456E7375726541646D696E4F6E6C79005365636F6E64617279007365636F6E64617279005072696D617279007072696D617279006F705F457175616C697479006F705F496E657175616C6974790000295300650063006F006E00640061007200790041006400640072006500730073003A007B0030007D0000174D0061007000700069006E0067003A007B0030007D00001F4D0061007000700069006E00670049006E0066006F003A007B0030007D00000B4F0077006E0065007200001D41007300730065007200740020006600610069006C00650064002E0000414D0061007000700069006E00670020006900730020006E006F007400200069006E002000700065006E00640069006E0067002000730074006100740065002E00003954006800650020006D0061007000700069006E00670020006900730020006E006F007400200061007000700072006F007600650064002E00006754006800650020007B0030007D00200070006100720061006D0065007400650072002000630061006E0020006E006F0074002000620065002000640065006600610075006C00740028007A00650072006F002900200061006400640072006500730073002E00000B6F0077006E0065007200003F4F006E006C007900200063006F006E007400720061006300740020006F0077006E00650072002000630061006E0020006100630063006500730073002E0000000000EEE2C7F974D6C54DB86E6FFA481534DB00042001010803200001052001011111042000122D0500020E0E1C05200111190E062002010E1119063001011E000E040A01111007300102010E1E00042001010E05200101121D040701111005200201020E0420001239042000111908070311101110111406300101011E00040A011114030611190700020211191119087CEC85D7BEA7798E0400000000040100000004020000000206080306110C06200111191119072002011119111906200111101119072002011119111005200101111907200201121D111904280011190801000800000000001E01000100540216577261704E6F6E457863657074696F6E5468726F777301080100020000000000000000000000000000000000000010000000000000000000000000000000982C00000000000000000000B22C0000002000000000000000000000000000000000000000000000A42C0000000000000000000000005F436F72446C6C4D61696E006D73636F7265652E646C6C0000000000FF2500200010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002000000C000000C43C00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
```

PR #27, opened 2020-12-31
[Mainnet] Dividend Token Contract

YakupIpek:
# Dividend Token Contract

**Compiler**
```
v0.0.2
```

**Contract Hash**
```
afb554ad1bb27c85f79f9ed061bd28e4cacbe5fb6fd29231c57f2047ebe00151
```

**Contract Byte Code**
```
4D5A90000300000004000000FFFF0000B800000000000000400000000000000000000000000000000000000000000000000000000000000000000000800000000E1FBA0E00B409CD21B8014CCD21546869732070726F6772616D2063616E6E6F742062652072756E20696E20444F53206D6F64652E0D0D0A2400000000000000504500004C0102004E678B9A0000000000000000E00022200B01300000140000000200000000000052320000002000000040000000000010002000000002000004000000000000000400000000000000006000000002000000000000030040850000100000100000000010000010000000000000100000000000000000000000003200004F000000000000000000000000000000000000000000000000000000004000000C000000E43100001C0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000200000080000000000000000000000082000004800000000000000000000002E7465787400000058120000002000000014000000020000000000000000000000000000200000602E72656C6F6300000C00000000400000000200000016000000000000000000000000000040000042000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000034320000000000004800000002000500982500004C0C000001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004602280600000A72010000706F0700000A2A4A02280600000A7201000070036F0800000A2A7202280600000A7215000070038C08000001280900000A6F0100002B2A7602280600000A7215000070038C08000001280900000A046F0200002B2AC20203280C00000A0204281500000602052813000006020E0428110000060202280D00000A6F0E00000A0428170000062A660202280100000602280D00000A6F0F00000AD728020000062A8E0202280D00000A6F0E00000A280900000626020328090000062602030428180000062A6A020328090000062602042809000006260203040528190000062A0000133003003C00000001000011020328030000060A020306280A0000060B07166A36241200067B0100000407D77D0100000412000228010000067D030000040203062804000006062A133002002100000002000011022801000006047B03000004DB0A0203281600000606D90B047B0100000407D72A4A0202280D00000A6F0E00000A280C0000062A133003001800000003000011020328030000060A020306280A0000060228140000065C2A4A0202280D00000A6F0E00000A280E0000062A00133003001F00000003000011020328030000060A020306280A000006067B02000004D70228140000065C2A001330030093000000040000110202280D00000A6F0E00000A28090000060A067B010000040228140000065C0B067B010000040228140000065E0C0207166AFE03722D000070281000000A1200067B02000004067B01000004D708DB7D020000041200087D010000040202280D00000A6F0E00000A0628040000060202280D00000A6F0E00000A07281100000A0D02096F1200000A7269000070281000000A2A4602280600000A728B0000706F1300000A2A4A02280600000A728B000070036F1400000A2A4602280600000A72990000706F1300000A2A4A02280600000A7299000070036F1400000A2A4602280600000A72A30000706F0700000A2A4A02280600000A72A3000070036F0800000A2A7202280600000A72BB000070038C08000001280900000A6F0700000A2A7602280600000A72BB000070038C08000001280900000A046F0800000A2A00000013300400A600000005000011042D34021201FE1504000002120102280D00000A6F0E00000A7D040000041201037D050000041201166A7D0600000407280300002B172A0202280D00000A6F0E00000A28160000060A06043402162A0202280D00000A6F0E00000A0604DB281700000602030203281600000604D72817000006021201FE1504000002120102280D00000A6F0E00000A7D040000041201037D050000041201047D0600000407280300002B172A000013300500AA00000006000011052D2A021202FE15040000021202037D040000041202047D050000041202166A7D0600000408280300002B172A020302280D00000A6F0E00000A281C0000060A020328160000060B0605370407053402162A020302280D00000A6F0E00000A0605DB281B00000602030705DB281700000602040204281600000605D72817000006021202FE15040000021202037D040000041202047D050000041202057D0600000408280300002B172A00001330040065000000070000110202280D00000A6F0E00000A03281C000006042E02162A0202280D00000A6F0E00000A0305281B000006021200FE1505000002120002280D00000A6F0E00000A7D070000041200037D080000041200057D0A0000041200047D0900000406280400002B172A8E02280600000A72D3000070038C08000001048C08000001281600000A056F0800000A2A8A02280600000A72D3000070038C08000001048C08000001281600000A6F0700000A2A42534A4201000100000000000C00000076342E302E33303331390000000005006C00000030050000237E00009C0500004804000023537472696E677300000000E4090000F800000023555300DC0A0000100000002347554944000000EC0A00006001000023426C6F6200000000000000020000015717A201090A000000FA0133001600000100000010000000050000000A0000001C0000002500000001000000160000000800000007000000010000000400000008000000010000000300000003000000040000000000A3010100000000000600EE001D0306002D011D030600DA0070020F003D0300000A001D0168030A00860368030E000C02FC020A00580368030A009B0068030A00AF03680306009100D2010A000E0168030A00AF00680306007801D20106009403D2010A0065006803000000001500000000000100010001001000FE0100001900010001000A011000DB0300002D0001001D000A0110008B0100002D0004001D000A0110007F0100002D0007001D0006003500B7000600DC02B70006009F02B7000600F401BA0006003402BA000600FC03B70006005E02BA0006004502BA000600EB03B7000600FC03B700502000000000860883024A00010062200000000081089102BE0001007520000000008100CD03C30002009220000000008100D803CA000300B0200000000086186A02D2000500E12000000000C6005B0106000900FB2000000000E6011B02DB0009001F2100000000E601D901E2000B003C21000000008100BF03C3000E008421000000008100B102EB000F00B121000000008600EF024A001100C421000000008600EF02F3001100E821000000008600CA024A001200FC21000000008600CA02F30012002822000000008600110406001300C722000000008608B501F9001300D922000000008108C001FD001300EC220000000086087A00F9001400FE220000000081088300FD001400112300000000E6091A044A00150023230000000081082A04BE001500362300000000E6014500F3001600532300000000810050000201170074230000000081002602DB0019002824000000008100E601E2001B00E02400000000E601630109011E005125000000008100970111012100752500000000E6015B001A01240000000100550100000100600300000100600300000200E30300000100D400000002003A04000003008C0000000400CB0100000100370200000200030400000100F90100000200370200000300030400000100600300000100600300000200E30300000100600300000100600300000100550100000100550100000100550100000100600300000100600300000200550100000100370200000200030400000100F901000002003702000003000304000001004D0200000200F503000003000304000001006402000002004D02000003005501000001006402000002004D0202001D0009006A02010011006A02060019006A020A0029006A02060061006A0206003100C000100069000100150069000B001A0071007F03200069009B0326006900A503320031006A023A0031006E00400081003A02450081004B014A0031000A046700310055026D0051004C03750069006B017900690075017E00310093018A0071007F03A7002E000B002A012E00130033012E001B005201430023005B0181002B005B01A1002B005B01E1002B005B0101012B005B014E00540059005E00840096009D00020001000000F20222010000C401260100008700260100002E04220102000100030001000200030002001000050001001100050002001200070001001300070002001400090001001500090004800000000000000000000000000000000086030000040000000000000000000000AE001E000000000001000200010000000000000000006803000000000100000000000000000000000000FC020000000003000200040002000500020015002D0017002D002B0091002B00A20000000047657455496E7436340053657455496E743634003C4D6F64756C653E0053797374656D2E507269766174652E436F72654C6962004469766964656E6442616C616E63650047657442616C616E63650053657442616C616E636500416C6C6F77616E636500494D657373616765006765745F4D657373616765006765745F4E616D65007365745F4E616D65006E616D650056616C7565547970650049536D617274436F6E74726163745374617465004950657273697374656E745374617465006765745F50657273697374656E7453746174650073746174650044656275676761626C6541747472696275746500436F6D70696C6174696F6E52656C61786174696F6E7341747472696275746500496E646578417474726962757465004465706C6F794174747269627574650052756E74696D65436F6D7061746962696C697479417474726962757465006765745F56616C75650076616C7565005265636569766500417070726F766500476574537472696E6700536574537472696E6700417070726F76616C4C6F67005472616E736665724C6F6700536574417070726F76616C00536D617274436F6E74726163742E646C6C006765745F53796D626F6C007365745F53796D626F6C0073796D626F6C0053797374656D005472616E7366657246726F6D005472616E73666572546F6B656E7346726F6D0066726F6D004469766964656E64546F6B656E00495374616E64617264546F6B656E005472616E73666572546F005472616E73666572546F6B656E73546F00746F006765745F53656E646572005370656E646572007370656E646572005472616E73666572004F776E6572006F776E6572002E63746F720053797374656D2E446961676E6F7374696373006765745F4469766964656E6473007365745F4469766964656E64730043726564697465644469766964656E647300476574576974686472617761626C654469766964656E647300476574546F74616C4469766964656E64730057697468647261776E4469766964656E6473004765744469766964656E647300537472617469732E536D617274436F6E7472616374732E5374616E64617264730053797374656D2E52756E74696D652E436F6D70696C6572536572766963657300446562756767696E674D6F646573006765745F537563636573730041646472657373006164647265737300537472617469732E536D617274436F6E74726163747300466F726D617400536D617274436F6E7472616374004F626A656374004765745374727563740053657453747275637400495472616E73666572526573756C74005570646174654163636F756E74004765744163636F756E74005365744163636F756E74006163636F756E74004F6C64416D6F756E740063757272656E74416D6F756E7400616D6F756E7400417373657274005769746864726177006765745F546F74616C537570706C79007365745F546F74616C537570706C7900746F74616C537570706C7900000000134400690076006900640065006E006400730000174100630063006F0075006E0074003A007B0030007D00003B54006800650020006100630063006F0075006E007400200068006100730020006E006F0020006400690076006900640065006E00640073002E0000215400720061006E00730066006500720020006600610069006C00650064002E00000D530079006D0062006F006C0000094E0061006D006500001754006F00740061006C0053007500700070006C0079000017420061006C0061006E00630065003A007B0030007D00002341006C006C006F00770061006E00630065003A007B0030007D003A007B0031007D00000031EC77F44ED48A46A24645E9865F8F930004200101080320000105200101111104200012350420010B0E052002010E0B0500020E0E1C063001011E000E040A01110C07300102010E1E00052001011225042000124104200011210320000B050702110C0B0407020B0B040701110C080704110C0B0B122905200201020E072002122911210B032000020420010E0E052002010E0E0507020B111006300101011E00040A0111100607030B0B11100407011114040A0111140600030E0E1C1C087CEC85D7BEA7798E02060B03061121042001010B062001110C1121072002011121110C0820040112250B0E0E0620020211210B08200302112111210B0720020B1121110C0520010B11210320000E042001010E0620020111210B0720030211210B0B08200301112111210B0720020B112111210328000B0328000E0801000800000000001E01000100540216577261704E6F6E457863657074696F6E5468726F7773010801000200000000000401000000000000000000000000000000100000000000000000000000000000002832000000000000000000004232000000200000000000000000000000000000000000000000000034320000000000000000000000005F436F72446C6C4D61696E006D73636F7265652E646C6C0000000000FF250020001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000003000000C000000543200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
```


Mainnet/DividendToken/DividendToken/DividendToken.cs
```diff
@@ -0,0 +1,288 @@
...
+public class DividendToken : SmartContract, IStandardToken
+{
+
+    public ulong Dividends
+    {
+        get => PersistentState.GetUInt64(nameof(this.Dividends));
+        private set => PersistentState.SetUInt64(nameof(this.Dividends), value);
+    }
+
+    private Account GetAccount(Address address) => PersistentState.GetStruct<Account>($"Account:{address}");
```
mrtpain, line 14: Should this be private for any particular reason?

Mainnet/DividendToken/DividendToken/DividendToken.cs
```diff
@@ -0,0 +1,288 @@
...
+
+    private void SetAccount(Address address, Account account) => PersistentState.SetStruct($"Account:{address}", account);
+
+
+    public DividendToken(ISmartContractState state, ulong totalSupply, string name, string symbol)
```
mrtpain, line 19: Per Issue https://github.com/stratisproject/CirrusSmartContracts/issues/29, I believe `totalSupply` and other token value related fields should be updated with this issue proposal to match ETH's solidity `UInt256` token based values for standardization amongst token types and wrapped Eth tokens.

Mainnet/DividendToken/DividendToken/DividendToken.cs
```diff
@@ -0,0 +1,288 @@
...
+        : base(state)
+    {
+        this.TotalSupply = totalSupply;
+        this.Name = name;
+        this.Symbol = symbol;
+        this.SetBalance(Message.Sender, totalSupply);
+    }
```
mrtpain, line 26: `Decimals` also need to be added to IStandrdToken interface and set in the constructor of all new tokens. This is for display purposes so clients of the contract know how to properly display satoshi values.

Existing PR for the Decimal value in StandardToken contract here: https://github.com/stratisproject/CirrusSmartContracts/pull/17

Still needs a PR to IStandardToken interface in Stratis.SmartContracts.Standards

Mainnet/DividendToken/DividendToken/DividendToken.cs
```diff
@@ -0,0 +1,288 @@
...
+
+    /// <summary>
+    /// It is advised that deposit amount should to be evenly divided by total supply, 
+    /// otherwise small amount of satoshi may lost(burn)
+    /// </summary>
+    public override void Receive()
+    {
+        Dividends += Message.Value;
```
mrtpain, line 34: Using `checked()` might be a good idea here.

Mainnet/DividendToken/DividendToken/DividendToken.cs
```diff
@@ -0,0 +1,288 @@
...
+
+    public bool TransferFrom(Address from, Address to, ulong amount)
+    {
+        UpdateAccount(from);
+        UpdateAccount(to);
+
+        return TransferTokensFrom(from, to, amount);
+    }
+
+    Account UpdateAccount(Address address)
```
mrtpain, line 53: Add `public` or `private`

Mainnet/DividendToken/DividendToken/DividendToken.cs
```diff
@@ -0,0 +1,288 @@
...
+    {
+        var account = GetAccount(address);
```
mrtpain, line 55: Check that the account actually exists and is valid.

Mainnet/DividendToken/DividendToken/DividendToken.cs
```diff
@@ -0,0 +1,288 @@
...
+        var newDividends = GetWithdrawableDividends(address, account);
+
+        if (newDividends > 0)
+        {
+            account.DividendBalance = checked(account.DividendBalance + newDividends);
+            account.CreditedDividends = Dividends;
```
mrtpain, line 61: The CreditedDividends for this account should equal the entire value of all `Dividends`?

Mainnet/DividendToken/DividendToken/DividendToken.cs
```diff
@@ -0,0 +1,288 @@
...
+            SetAccount(address, account);
+        }
+
+        return account;
+    }
+
+    private ulong GetWithdrawableDividends(Address address, Account account)
+    {
+        var newDividends = Dividends - account.CreditedDividends;
+        var notCreditedDividends = checked(GetBalance(address) * newDividends);
```
mrtpain, line 71: Partial reason for using `UInt256` for token value properties right here. Huge chance of overflow exception thrown multiplying `ulong * ulong = ulong`. This can be a critical issue.

mrtpain (changes requested):
Requesting changes be made due to `GetWithdrawableDividends` methods multiplication. 
https://github.com/stratisproject/CirrusSmartContracts/pull/27/files#r550697142

If you write tests for this and every day real scenario values, you will find overflow exceptions very easily breaking the contract.

Mainnet/DividendToken/DividendToken/DividendToken.cs
mrtpain: For added context, max ulong value is `18446744073709551615`, square root of that is `4294967296` which is equal to `42.94967296` tokens. Making any combination of values higher than that multiplied by each other result in the overflow exception metioned.

Mainnet/DividendToken/DividendToken/DividendToken.cs
YakupIpek: Dividends  value can reach to max total supply of Cirrus token which is lower than ulong max value

Mainnet/DividendToken/DividendToken/DividendToken.cs
YakupIpek: Account data is actually internal purposes only and properties of Account is not in final state. They should to be computed. The needed methods already exposed on contract as `GetDividends` and `GetTotalDividends` . Client can also calculate Withdrawn dividiends by computing value of `GetTotalDividend - GetDividends `

Mainnet/DividendToken/DividendToken/DividendToken.cs
YakupIpek: Private keyword added

Mainnet/DividendToken/DividendToken/DividendToken.cs
YakupIpek: It will return default value for Account struct which is expected behavior. Account value can't be invalid.

Mainnet/DividendToken/DividendToken/DividendToken.cs
YakupIpek: I may update the code when that #17 approved but i will stick to current implementation of StandardToken for now.

Mainnet/DividendToken/DividendToken/DividendToken.cs
YakupIpek: Max value of `newDividends` can be Total Supply of Cirrus coin which is 100 million and max value of `GetBalance(address)` can be Total Supply of dividend token so in this case, it can overflow if total supply is higher than 184.467.440.737 number.  It is less likely that total supply can be such big number but it  can be.  We may restrict total supply for higher than 184.467.440.737 number.

Mainnet/DividendToken/DividendToken/DividendToken.cs
YakupIpek: I know it looks weird at first sight without knowing design decisions. I added description but it is not covering technical design decisions. 

Dividends incremented whenever contract get deposits by `DistributeDividends` or `Receive` methods.  But we don't calculate  address deposits in those methods because it requires iteration on all token holder addresses which is very costly and iterative code is bad design for smart contracts. The contract lazily compute dividends whenever tokens transferred By Transfer methods.  So each address tracks its own computed Dividends by `CreditedDividends`  If `Dividends == CreditedDividends`  then all deposited dividends already computed and added `DividendBalance`  or If it is not equal so there are some new dividends (`Dividends - CreditedDividends`) which is not yet computed. 


As you may see `DividendBalance`  should be divided by `TotalSupply` but it is not. Because it is saving divided value cause precision loss. So we save values without divide operation but we always display dividends by dividing to TotalSupply when user request it.

Mainnet/DividendToken/DividendToken/DividendToken.cs
mrtpain: I do not suggested pushing out a smart contract that has these limitations or possible overflow exceptions. Even if the chance is small, it will lock tokens from being able to be withdrawn and result in loss of funds.

Mainnet/DividendToken/DividendToken/DividendToken.cs
mrtpain: Looks like that PR is up to you and I to approve and merge, nobody else is looking at it and it's been there for months. The change matches ERC token standards and is needed for Opdex and general wallets for token display purposes. Not a good idea to have mainnet tokens without it.

Mainnet/DividendToken/DividendToken/DividendToken.cs
```diff
@@ -116,7 +115,8 @@ public ulong GetDividends(Address address)
     public ulong GetTotalDividends(Address address)
     {
         var account = GetAccount(address);
-        return checked(GetWithdrawableDividends(address, account) + account.WithdrawnDividends) / TotalSupply;
+        var withdrawable = GetWithdrawableDividends(address, account) / TotalSupply;
+        return withdrawable + account.WithdrawnDividends;
```
mrtpain: Are the latest updates correcting the previous code here? 

Initially had `(a + b) / c`
Updated to `(a / c) + b`

mrtpain (approved):
One comment/question otherwise lgtm

Mainnet/DividendToken/DividendToken/DividendToken.cs
YakupIpek: If you look changes you can see that b value is divided to c still and saved to persistent state. It is for avoid being uint256 type because b/c is specify CRS token amount which is ulong.

PR #28, opened 2020-12-31
[Mainnet] Non Fungible Token Contract

YakupIpek:
# Non-Fungible Token Token Contract

**Compiler Version**
```
v0.0.2
```

**Contract Hash**
```
28ac3374260071c654beaa9c4587304428061c9b7c3d0318e6ae33fc932f21d5
```

**Contract Byte Code**
```
4D5A90000300000004000000FFFF0000B800000000000000400000000000000000000000000000000000000000000000000000000000000000000000800000000E1FBA0E00B409CD21B8014CCD21546869732070726F6772616D2063616E6E6F742062652072756E20696E20444F53206D6F64652E0D0D0A2400000000000000504500004C010200557658CE0000000000000000E00022200B013000002200000002000000000000AA410000002000000060000000000010002000000002000004000000000000000400000000000000008000000002000000000000030040850000100000100000000010000010000000000000100000000000000000000000584100004F000000000000000000000000000000000000000000000000000000006000000C0000003C4100001C0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000200000080000000000000000000000082000004800000000000000000000002E74657874000000B0210000002000000022000000020000000000000000000000000000200000602E72656C6F6300000C0000000060000000020000002400000000000000000000000000004000004200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000008C410000000000004800000002000500942A0000A816000001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000007202280500000A7201000070038C0C000001280600000A6F0700000A2A7602280500000A7201000070038C0C000001280600000A046F0800000A2A46722F000070038C0F000001280600000A2A4E02280500000A020328030000066F0900000A2A5202280500000A02032803000006046F0A00000A2A46724B000070038C0F000001280600000A2A4E02280500000A020328060000066F0900000A2A5202280500000A02032806000006046F0A00000A2A7202280500000A726D000070038C06000001280600000A6F0B00000A2A7602280500000A726D000070038C06000001280600000A046F0C00000A2A8A02280500000A729D000070038C06000001048C06000001280D00000A6F0700000A2A8E02280500000A729D000070038C06000001048C06000001280D00000A056F0800000A2A4602280500000A72CD0000706F0900000A2A4A02280500000A72CD000070036F0A00000A2A4602280500000A72D90000706F0E00000A2A4A02280500000A72D9000070036F0F00000A2A4602280500000A72E30000706F0E00000A2A4A02280500000A72E3000070036F0F00000A2A4602280500000A72F10000706F0B00000A2A4A02280500000A72F1000070036F0C00000A2A467209010070038C0F000001280600000A2A4E02280500000A020328150000066F0B00000A2A5202280500000A02032815000006046F0C00000A2A4E02280500000A020328150000066F1000000A2A46722B010070038C0F000001280600000A2A4E02280500000A020328190000066F0B00000A2A5202280500000A02032819000006046F0C00000A2A4E02280500000A020328190000066F1000000A2A5E724D010070038C06000001048C0F000001280D00000A2A5202280500000A020304281D0000066F0B00000A2A5602280500000A020304281D000006056F0C00000A2A5202280500000A020304281D0000066F1000000A2A5E7285010070038C06000001048C0F000001280D00000A2A5202280500000A02030428210000066F0B00000A2A5602280500000A0203042821000006056F0C00000A2A5202280500000A02030428210000066F1000000A2A4602280500000A72BD0100706F0B00000A2A4A02280500000A72BD010070036F0C00000A2A00001330030057000000000000000203281100000A021717280200000602181728020000060219162802000006021A172802000006021B17280200000602042810000006020528120000060202281200000A6F1300000A280E00000602176A28140000062A720203022825000006FE0572D5010070281400000A020328160000062A7A020402032809000006FE0572D5010070281400000A020304281E0000062A22020328010000062A32020304050E0428370000062A4202030405168D1100000128370000062A00001330030038000000010000110205283D000006020528040000060A0206284300000602042843000006020603281500000A7201020070281400000A02040528340000062A133004003A000000010000110204283C0000060204283E000006020428040000060A020306281600000A7201020070281400000A020403280800000602060304283A0000062A9E0202281200000A6F1300000A0304280C0000060202281200000A6F1300000A0304283B0000062A3E02032843000006020328090000062A0000133002001100000001000011020328040000060A02062843000006062A3E0203283E000006020328070000062A26020304280B0000062A00133004002900000001000011020428040000060A02042838000006020604283500000602030428360000060206030428390000062A000000133004007B00000002000011020204280400000603281500000A7201020070281400000A020328090000060A020306176ADB280A00000602280500000A020428030000066F1000000A02030428220000060B06176ADB0C07082E1B020308281E0000060D02030907282300000602030709281F000006020308282000000602030428240000062A00133004004C0000000300001102020428040000067E1700000A281500000A7201020070281400000A0204032805000006020328090000060A020306176AD7280A000006060B02030407282300000602030704281F0000062A133008009D000000040000110205283D0000060205283E000006020528040000060A020603281500000A7201020070281400000A02042843000006020405283400000602280500000A046F1800000A2C570204166A721F0200701A8D0E000001251602281200000A6F1300000A8C06000001A22517038C06000001A22518058C0F000001A225190E04A2166A281900000A0B02076F1A00000AA5120000017201020070281400000A2A9A020328070000067E1700000A281600000A2C1202280500000A020328060000066F1000000A2A133003002800000005000011021200FE15030000021200037D010000041200047D020000041200057D0300000406280100002B2A133003002800000006000011021200FE15040000021200037D040000041200047D050000041200057D0600000406280200002B2A133003002800000007000011021200FE15050000021200037D070000041200047D080000041200057D0900000406280300002B2A133004003C00000001000011020328040000060A020602281200000A6F1300000A281500000A2D14020602281200000A6F1300000A280B0000062B01177201020070281400000A2A133004005500000001000011020328040000060A020602281200000A6F1300000A281500000A2D2D0203280700000602281200000A6F1300000A281500000A2D14020602281200000A6F1300000A280B0000062B01177201020070281400000A2A000000133002001000000001000011020328040000060A020628430000062A13300400520000000800001102284000000602037E1700000A281600000A725502007072BD020070280600000A281400000A021200FE1506000002120002280D0000067D0A0000041200037D0B00000406280400002B0203280E0000062A8A0202281200000A6F1300000A02280D000006281500000A72C9020070281400000A2A000000133003009800000009000011022840000006020328430000060204166AFE037225030070281400000A0228250000060A0604D70B0228130000060C2B4D02030828360000060206082817000006020806281B000006021203FE150300000212037E1700000A7D010000041203037D020000041203087D0300000409280100002B06176AD70A08176AD70C060737AF0202282500000604D72826000006020828140000062A13300300AB0000000A000011020328040000060A02062843000006020602281200000A6F1300000A281500000A7271030070281400000A0203283800000602060328350000060203281A0000060B022825000006176ADB1304021104282600000611040C020828160000060D0207092817000006020907281B000006020828180000060203281C000006021205FE15030000021205067D0100000412057E1700000A7D020000041205037D030000041105280100002B2A5E02037E1700000A281600000A72BB030070281400000A2A0042534A4201000100000000000C00000076342E302E33303331390000000005006C0000004C090000237E0000B80900001C07000023537472696E677300000000D4100000F803000023555300CC140000100000002347554944000000DC140000CC01000023426C6F6200000000000000020000015715A201090A000000FA0133001600000100000012000000060000000B000000430000006A0000001B0000000D0000000A00000001000000050000000A000000010000000200000004000000040000000000BC0201000000000006003F01D90406006E01D90406002B0198040F00F90400000A005F0536050A00260536050A00EC0036050A00740536050600D700FB020A005F0136050A000001360506000100FB020600CE01FB0206006D05FB0206001500FB020A00AB00360506008C01FB0206001803FB02000000001C00000000000100010001001000460300001500010001000A0110000A0200002500010044000A011000EC0100002500040044000A011000F80100002500070044000A011000D501000025000A00440006000E0384000600DB03840006007100F00006006004840006008400840006007100F00006006004840006008904840006008400F30006004F04840006005D0484005020000000008100AB04F60001006D20000000008100C204FB0002008B200000000081007C06010104009D20000000008100350406010500B12000000000810042040C010600C6200000000081003F0601010800D8200000000081004C0206010900EC200000000081005C020C010A000121000000008100840513010C001E210000000081009B0519010D003C210000000081006C0420010F005F210000000081007F0428011100832100000000860821046000140095210000000081082B0431011400A821000000008608C00037011500BA21000000008108C90050001500CD21000000008608CE0237011600DF21000000008108D90250001600F22100000000810859003B011700042200000000810869003F01170017220000000081008C06010118002922000000008100D205440119003D22000000008100E20549011A005222000000008100C0053F011C006622000000008100690601011D007822000000008100B00344011E008C22000000008100C00349011F00A1220000000081009E033F012100B5220000000081009F064F012200CD220000000081000B0656012400E22200000000810022065D012600F822000000008100F205190129000D2300000000810052064F012B002523000000008100700356012D003A2300000000810087035D012F0050230000000081005703190132006523000000008608C9063B0134007723000000008108D9063F0134008C23000000008618920465013500EF23000000008600E505440138000C240000000086002506560139002B240000000086009900F6003B00342400000000860002036D013C00412400000000860002037801400054240000000086000603780143009824000000008600A70119014600DE240000000086009D02810148000625000000008600AF0113014A001825000000008600B90106014B003525000000008600810006014C0045250000000086007A0220014D0050250000000081002F0219014F0088250000000081003803190151001026000000008100200319015300682600000000810016026D01550011270000000081006C023F0159003827000000008100090478015A006C27000000008100400278015D00A0270000000081008B0228016000D427000000008100E1003F0163001C2800000000810015043F01640080280000000081002B033F0165009C28000000008600E60331016600FA28000000008100B906060067002029000000008600AF0219016700C429000000008600D6033F0169007B2A000000008600030731016A00000001004D00000001004D0000000200A10100000100960000000100960000000100960000000200A10100000100960000000100960000000100960000000200A101000001002E05000001002E0500000200A10100000100660400000200080500000100660400000200080500000300A10100000100A10100000100A10100000100A10100000100A10100000100390600000100390600000100390600000200D00300000100390600000100D00300000100D00300000100D00300000200390600000100D003000001002E05000002003906000001002E0500000200390600000100660400000200390600000300790000000100660400000200390600000100660400000200790000000100660400000200790000000100660400000200790000000300390600000100660400000200790000000100A10100000100250100000200D20000000300E40200000100390600000100660400000200390600000100250000000100130300000200E30300000300790000000400310000000100130300000200E30300000300790000000100130300000200E303000003007900000001008D00000002007900000001000805000002008D0000000100660400000100790000000100790000000100660400000200080500000100E30300000200790000000100130300000200790000000100E30300000200790000000100130300000200E30300000300790000000400310000000100790000000100130300000200E303000003007900000001006604000002008D00000003007900000001006604000002000805000003008D00000001007900000001007900000001007900000001006604000001002E0500000200B205000001007900000001002E05090092040100110092040600190092040A0051009204060029001101100069004D0515005900EB021B005900F3022000590018052600590023052C0059000800330059001200380069004D053E005900C10145005900CB014A005900F80350002900920455002900B4005B008100FE0360002900B90565003100E90670003100F50670003100DE038400590054058F002900B702950041009101A10029001202AA0021002300C6012E000B0095012E0013009E012E001B00BD0141002300C60161002300C60181002300C601A1002300C601C1002300C601E1002300C60101012300C60141012300C60161012300C6016B0078007F008800A500B600C000CA00D400DC00020001000000600488010000CD008D010000DD028D0100006D0091010000DD06910102000D00030001000E00030002000F000500010010000500020011000700010012000700020013000900010014000900020025000B00010026000B000480000000000000000000000000000000005F050000040000000000000000000000E7003600000000000100020001000000000000000000360500000000030002000400020005000200060002003700B1003700BB003700C5003700CF0000000055496E7433320047657455496E7436340053657455496E743634003C4D6F64756C653E00696E74657266616365494400646174610053797374656D2E507269766174652E436F72654C696200696E746572666163654964006765745F4E657874546F6B656E4964007365745F4E657874546F6B656E496400746F6B656E496400476574417070726F76656400617070726F76656400696400537570706F727473496E7465726661636500494D657373616765006765745F4D657373616765006765745F4E616D65007365745F4E616D65006E616D650056616C7565547970650043616E4F7065726174650049536D617274436F6E74726163745374617465004950657273697374656E745374617465006765745F50657273697374656E7453746174650073746174650044656275676761626C6541747472696275746500436F6D70696C6174696F6E52656C61786174696F6E7341747472696275746500496E6465784174747269627574650052756E74696D65436F6D7061746962696C6974794174747269627574650042797465006765745F52657475726E56616C75650076616C756500417070726F76650042616C616E63654F66004F776E65724F6600476574537472696E6700536574537472696E67004F776E6572736869705472616E7366657265644C6F6700417070726F76616C4C6F6700417070726F76616C466F72416C6C4C6F67005472616E736665724C6F6700536166655472616E7366657246726F6D496E7465726E616C005472616E73666572496E7465726E616C004C6F67417070726F76616C004765744964546F417070726F76616C005365744964546F417070726F76616C00436C656172417070726F76616C004973417070726F766564466F72416C6C004C6F67417070726F76616C466F72416C6C00536574417070726F76616C466F72416C6C004D696E74416C6C0043616C6C00536D617274436F6E74726163742E646C6C006765745F53796D626F6C007365745F53796D626F6C0073796D626F6C00476574426F6F6C00536574426F6F6C0053797374656D00536166655472616E7366657246726F6D0066726F6D00426F6F6C65616E004164644E46546F6B656E0056616C69644E46546F6B656E0052656D6F76654E46546F6B656E004E6F6E46756E6769626C65546F6B656E00436C656172496E6465784F664F776E65724279546F6B656E00476574496E6465784F664F776E65724279546F6B656E00536574496E6465784F664F776E65724279546F6B656E00436C656172496E6465784279546F6B656E00476574496E6465784279546F6B656E00536574496E6465784279546F6B656E00746F6B656E004275726E00546F005A65726F00746F005472616E736665724F776E65727368697000436C656172006765745F53656E646572004C6F675472616E736665720043616E5472616E73666572006765745F4F776E6572007365745F4F776E6572004765744964546F4F776E6572005365744964546F4F776E65720050726576696F75734F776E6572004E65774F776E6572006F776E6572004765744F776E6572546F4F70657261746F72005365744F776E6572546F4F70657261746F72002E63746F720053797374656D2E446961676E6F737469637300476574537570706F72746564496E746572666163657300536574537570706F72746564496E74657266616365730053797374656D2E52756E74696D652E436F6D70696C6572536572766963657300446562756767696E674D6F646573006F70657261746F724164647265737300476574416464726573730053657441646472657373006164647265737300537472617469732E536D617274436F6E74726163747300466F726D6174004973436F6E747261637400536D617274436F6E7472616374004F626A65637400495472616E73666572526573756C74004765744F776E6572546F4E46546F6B656E436F756E74005365744F776E6572546F4E46546F6B656E436F756E7400616D6F756E740041737365727400436C656172546F6B656E4279496E64657800476574546F6B656E4279496E64657800536574546F6B656E4279496E64657800436C656172546F6B656E4F664F776E65724279496E64657800476574546F6B656E4F664F776E65724279496E64657800536574546F6B656E4F664F776E65724279496E64657800696E646578004765744964546F417070726F76616C4B657900496E6465784F664F776E65724279546F6B656E4B657900476574496E6465784279546F6B656E4B6579004765744964546F4F776E65724B657900476574546F6B656E4279496E6465784B657900476574546F6B656E4F664F776E65724279496E6465784B657900456E737572654F776E65724F6E6C79006765745F546F74616C537570706C79007365745F546F74616C537570706C79006F705F457175616C697479006F705F496E657175616C69747900456E737572654164647265737349734E6F74456D7074790000002D53007500700070006F00720074006500640049006E0074006500720066006100630065003A007B0030007D00001B4900640054006F004F0077006E00650072003A007B0030007D0000214900640054006F0041007000700072006F00760061006C003A007B0030007D00002F4F0077006E006500720054006F004E00460054006F006B0065006E0043006F0075006E0074003A007B0030007D00002F4F0077006E006500720054006F004F00700065007200610074006F0072003A007B0030007D003A007B0031007D00000B4F0077006E006500720000094E0061006D006500000D530079006D0062006F006C0000174E0065007800740054006F006B0065006E0049006400002154006F006B0065006E004200790049006E006400650078003A007B0030007D00002149006E006400650078004200790054006F006B0065006E003A007B0030007D00003754006F006B0065006E004F0066004F0077006E00650072004200790049006E006400650078003A007B0030007D003A007B0031007D00003749006E006400650078004F0066004F0077006E00650072004200790054006F006B0065006E003A007B0030007D003A007B0031007D00001754006F00740061006C0053007500700070006C007900002B540068006500200069006E00640065007800200069007300200069006E00760061006C00690064002E00001D41007300730065007200740020006600610069006C00650064002E0000354F006E004E006F006E00460075006E006700690062006C00650054006F006B0065006E0052006500630065006900760065006400006754006800650020007B0030007D00200070006100720061006D0065007400650072002000630061006E0020006E006F0074002000620065002000640065006600610075006C00740028007A00650072006F002900200061006400640072006500730073002E00000B6F0077006E0065007200005B4F006E006C00790020006F0077006E006500720020006F0066002000740068006500200063006F006E00740072006100630074002000630061006E00200073006500740020006E006500770020006F0077006E00650072002E00004B740068006500200061006D006F0075006E0074002000730068006F0075006C006400200062006500200068006900670068006500720020007400680061006E0020007A00650072006F0000494F006E006C007900200074006F006B0065006E0020006F0077006E00650072002000630061006E0020006200750072006E002000740068006500200074006F006B0065006E002E000039540068006500200061006400640072006500730073002000630061006E0020006E006F00740020006200650020007A00650072006F002E0000000000F4E5AF4639F2AF489DE1629C30E7FAB300042001010803200001052001011111042000122D0500020E0E1C042001020E052002010E0205200111190E062002010E11190420010B0E052002010E0B0600030E0E1C1C0420010E0E052002010E0E042001010E05200101121D0420001241042000111905200201020E040701111907000202111911190607040B0B0B0B0407020B0B03061119060702111912210520010211190B2005122111190B0E1D1C0B0320001C040701110C06300101011E00040A01110C0407011110040A0111100407011114040A0111140407011118040A0111180707040B0B0B110C0A070611190B0B0B0B110C087CEC85D7BEA7798E02060B02060204200102090520020109020420010E0B05200111190B062002010B11190520010B11190620020111190B07200202111911190820030111191119020520010111190320000E0320000B042001010B0420010B0B052002010B0B0620020E11190B0620020B11190B0720030111190B0B07200301121D0E0E0A200401111911190B1D0508200301111911190B0620020111190204280011190328000E0328000B0801000800000000001E01000100540216577261704E6F6E457863657074696F6E5468726F777301080100020000000000040100000000000000000000000000000000100000000000000000000000000000008041000000000000000000009A4100000020000000000000000000000000000000000000000000008C410000000000000000000000005F436F72446C6C4D61696E006D73636F7265652E646C6C0000000000FF25002000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004000000C000000AC3100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
```


Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
```diff
@@ -0,0 +1,639 @@
...
+    /// </summary>
+    /// <param name="state">The smart contract state.</param>
+    public NonFungibleToken(ISmartContractState state, string name, string symbol) : base(state)
+    {
+        // todo: discuss callback handling and supported interface numbering with community.
+        this.SetSupportedInterfaces((uint)0x00000001, true); // (ERC165) - ISupportsInterface
+        this.SetSupportedInterfaces((uint)0x00000002, true); // (ERC721) - INonFungibleToken,
+        this.SetSupportedInterfaces((uint)0x00000003, false); // (ERC721) - INonFungibleTokenReceiver
+        this.SetSupportedInterfaces((uint)0x00000004, true); // (ERC721) - INonFungibleTokenMetadata
+        this.SetSupportedInterfaces((uint)0x00000005, true); // (ERC721) - IERC721Enumerable
```
mrtpain, line 216: Is there a practical use case for these interfaces? Normally in ETH you'd cast or tell the contract what interface another contract should be using. 

This sounds like ContractA interacting with this contract (Contract B) would have to make a call to check what interface is supported _before_ interacting with the contract. 

This is where whitelisted interfaces come into play.

ContractA should be able to say Assert(ContractB is ISomeInterface). I remember @Neurosploit I think it was that initially started this contract. Maybe he'll have some feedback.

Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
```diff
@@ -0,0 +1,639 @@
...
+
+    private string IndexOfOwnerByTokenKey(Address owner, ulong tokenId) => $"IndexOfOwnerByToken:{owner}:{tokenId}";
+    private ulong GetIndexOfOwnerByToken(Address owner, ulong tokenId) => this.PersistentState.GetUInt64(IndexOfOwnerByTokenKey(owner, tokenId));
+    private void SetIndexOfOwnerByToken(Address owner, ulong tokenId, ulong index) => this.PersistentState.SetUInt64(IndexOfOwnerByTokenKey(owner, tokenId), index);
+    private void ClearIndexOfOwnerByToken(Address owner, ulong tokenId) => this.PersistentState.Clear(IndexOfOwnerByTokenKey(owner, tokenId));
+    public ulong TotalSupply
+    {
+        get => this.PersistentState.GetUInt64(nameof(TotalSupply));
+        private set => this.PersistentState.SetUInt64(nameof(TotalSupply), value);
+    }
```
mrtpain, line 203: Same comment as other contracts, this should be changed for the better going forward as outlined in Issue https://github.com/stratisproject/CirrusSmartContracts/issues/29

Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
```diff
@@ -0,0 +1,639 @@
...
+            
+            checked
+            {
+                index++;
+                tokenId++;
+            }
+        }
+
+        TotalSupply = checked(TotalSupply + amount);
+        NextTokenId = tokenId;
```
mrtpain, line 607: `NextTokenId` is never beeing incremented. Line 589 - `var tokenId = NextTokenId` (initially 1 when contract is created). Then line 607 `NextTokenId = tokenId` should be `NextTokenId++` instead

Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
```diff
@@ -0,0 +1,639 @@
...
+
+        Address tokenOwner = GetIdToOwner(tokenId);
+        Assert(tokenOwner == from);
+        EnsureAddressIsNotEmpty(to);
+
+        TransferInternal(to, tokenId);
+
+        if (this.PersistentState.IsContract(to))
+        {
+            ITransferResult result = this.Call(to, 0, "OnNonFungibleTokenReceived", new object[] { this.Message.Sender, from, tokenId, data }, 0);
```
mrtpain, line 467: This method `OnNonFungibleTokenReceived` is in interfaces but is not implemented anywhere, where is this call supposed to be going to?

mrtpain (changes requested):
- I think we need to look into the use of Interfaces and real world uses with this contract.
- NextTokenId is never actually incremented
- OnNonFungibleTokenReceived is not a method anywhere
- Curious how much actual testing this has been done with and what the gas costs might be for some of these happy paths. There's a lot going on in a single action, thinking about 100,000 gas limits per transaction here, with callbacks / other contract integrations there needs to be sufficient room.

Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
Neurosploit: @mrtpain @YakupIpek  See the PR here https://github.com/stratisproject/StratisSmartContractsSamples/pull/7
That's as far as it has gotten and has left my memory completely so I can't be of much help.

Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
YakupIpek: `tokenId` is incremented at line 602

Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
YakupIpek: The contract which wants to be notified when they get a token(nft) should to implement that interface actually.  If NFT sends to a contract address then NFT contract notifies the client contract by doing external call.

Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
YakupIpek: Unfortunately `Assert(ContractB is ISomeInterface)` this type of casting is not yet possible. I guess early version of solidity didn't support it as well so that is why the original code include alternative way to call SupportInterface method.

Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
mrtpain: 😆 oops, missed that

Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
YakupIpek: Why `TotalSupply` should not stay as ulong ? It has enough large value for majority of use cases.

mrtpain: If the goal is to support wrapped ERC tokens, ulong won't be enough to support that since ethereum has up to 18 decimal places. `18.446744073709551615` would be the max total of WERC tokens in any of Stratis contracts

YakupIpek: This is not type of ERC20 token. These are assets which can not be divided, multiplied, substracted...etc. It can't have decimal point as well. So each token is corresponds to an asset. 

Original version is uint256 but i don't know why it is so .

YakupIpek: I am making research of reason it. But i guess this type of token is not going to use on dexes

Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
```diff
@@ -0,0 +1,639 @@
...
+        NextTokenId = tokenId;
+    }
+
+    public void Burn(ulong tokenId)
+    {
+        Address tokenOwner = GetIdToOwner(tokenId);
+
+        EnsureAddressIsNotEmpty(tokenOwner);
+
+        Assert(tokenOwner == Message.Sender, "Only token owner can burn the token.");
```
mrtpain, line 616: Message.Sender shouldn't ever be Address.Zero right? I think we can skip `EnsureAddressIsNotEmpty(tokenOwner)`

Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
```diff
@@ -0,0 +1,639 @@
...
+        var lastIndex = checked(index + amount);
+        var tokenId = NextTokenId;
+
+        while (index < lastIndex)
+        {
+            AddNFToken(address, tokenId);
+            SetTokenByIndex(index, tokenId);
+            SetIndexByToken(tokenId, index);
+
+            Log(new TransferLog { From = Address.Zero, To = address, TokenId = tokenId });
```
mrtpain, line 597: Can use the LogTransferMethod() here.

Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
```diff
@@ -0,0 +1,639 @@
...
+        var lastTokenIndex = checked(--TotalSupply);
+        var lastToken = GetTokenByIndex(lastTokenIndex);
+
+        SetTokenByIndex(index, lastToken);
+        SetIndexByToken(lastToken, index);
+
+        ClearTokenByIndex(lastTokenIndex);
+        ClearIndexByToken(tokenId);
+
+        Log(new TransferLog { From = tokenOwner, To = Address.Zero, TokenId = tokenId });
```
mrtpain, line 632: Can use the LogTransferMethod() here.

Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
```diff
@@ -0,0 +1,639 @@
...
+        //move last token to removed token and delete last token info
+        var index = GetIndexByToken(tokenId);
+        var lastTokenIndex = checked(--TotalSupply);
+        var lastToken = GetTokenByIndex(lastTokenIndex);
+
+        SetTokenByIndex(index, lastToken);
+        SetIndexByToken(lastToken, index);
+
+        ClearTokenByIndex(lastTokenIndex);
+        ClearIndexByToken(tokenId);
```
mrtpain, line 630: Should there be a log that a token has been moved to a different index? How would an owner know when their tokens index might have been changed due to a Burn?

Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
```diff
@@ -0,0 +1,639 @@
...
+    }
+
+    /// <summary>
+    /// Sets the contract owner who can mint/bur
+    /// </summary>
+    /// <param name="owner"></param>
+    public void TransferOwnership(Address owner)
+    {
+        EnsureOwnerOnly();
+        Assert(owner != Address.Zero, $"The {nameof(owner)} parameter can not be default(zero) address.");
```
mrtpain, line 564: Can probably use `EnsureAddressIsNotEmpty`

Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
mrtpain: should INonFungibleTokenReceiver always be false? Is there a reason to include it if so?

Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
mrtpain: Talked in discord but putting here for reference, I think you are right and uint64 will be ok for NFT's

Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
```diff
@@ -0,0 +1,639 @@
...
+
+    /// <summary>
+    /// Get a value indicacting if the interface is supported.
+    /// </summary>
+    /// <param name="interfaceId">The id of the interface to support.</param>
+    /// <returns>A value indicating if the interface is supported.</returns>
+    private bool GetSupportedInterfaces(uint interfaceId)
+    {
+        return this.PersistentState.GetBool($"SupportedInterface:{interfaceId}");
+    }
```
mrtpain, line 54: This Get method and others below it, should they be private? Would it be beneficial to make these public so external resources can read this data?

Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
YakupIpek: Done

Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
YakupIpek: done

Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
YakupIpek: Only owner can burn his own token already

Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
YakupIpek: There was public method for this already and i removed this one.

Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
YakupIpek: I will keep it as explicit documentation for now.

Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
mrtpain: Yes but then the `lastToken` is assigned to the index of the token that was just cleared and burned right? So if I'm the owner the the last token (index 99), and you burn your token (index 50), I get assigned to your old tokens index, how will I know? I'll still be calling the contract looking for index 99 token right?

Mainnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
```diff
@@ -0,0 +1,639 @@
...
+    private ulong GetTokenOfOwnerByIndex(Address address, ulong index) => this.PersistentState.GetUInt64(GetTokenOfOwnerByIndexKey(address, index));
+
+    private void SetTokenOfOwnerByIndex(Address owner, ulong index, ulong tokenId) => this.PersistentState.SetUInt64(GetTokenOfOwnerByIndexKey(owner, index), tokenId);
+
+    private void ClearTokenOfOwnerByIndex(Address owner, ulong index) => this.PersistentState.Clear(GetTokenOfOwnerByIndexKey(owner, index));
+
+    private string IndexOfOwnerByTokenKey(Address owner, ulong tokenId) => $"IndexOfOwnerByToken:{owner}:{tokenId}";
+    private ulong GetIndexOfOwnerByToken(Address owner, ulong tokenId) => this.PersistentState.GetUInt64(IndexOfOwnerByTokenKey(owner, tokenId));
+    private void SetIndexOfOwnerByToken(Address owner, ulong tokenId, ulong index) => this.PersistentState.SetUInt64(IndexOfOwnerByTokenKey(owner, tokenId), index);
+    private void ClearIndexOfOwnerByToken(Address owner, ulong tokenId) => this.PersistentState.Clear(IndexOfOwnerByTokenKey(owner, tokenId));
```
mrtpain: related to this comment: https://github.com/stratisproject/CirrusSmartContracts/pull/28#discussion_r565821159

I think all of the `Get` methods in this selection should be public.

YakupIpek: There are public methods already provide needed functionality. These are just getter type of methods properties and public method contains an extra logic over them.

PR #41, opened 2021-06-08
[Testnet] NFT Auction Store

YakupIpek:
# NFT Auction Store Contract

**Compiler Version**
```
v2.0.0
```
**Contract Hash**
```
bc12bedefb91d22cdd7318950cb1971825a26d2fa6afec04e56b18f11898b4c7
```

**Contract Byte Code**
```
4D5A90000300000004000000FFFF0000B800000000000000400000000000000000000000000000000000000000000000000000000000000000000000800000000E1FBA0E00B409CD21B8014CCD21546869732070726F6772616D2063616E6E6F742062652072756E20696E20444F53206D6F64652E0D0D0A2400000000000000504500004C010200D23F989D0000000000000000E00022200B013000001600000002000000000000CE3400000020000000400000000000100020000000020000040000000000000004000000000000000060000000020000000000000300408500001000001000000000100000100000000000001000000000000000000000007C3400004F000000000000000000000000000000000000000000000000000000004000000C000000603400001C0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000200000080000000000000000000000082000004800000000000000000000002E74657874000000D4140000002000000016000000020000000000000000000000000000200000602E72656C6F6300000C000000004000000002000000180000000000000000000000000000400000420000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000B0340000000000004800000002000500F0250000700E000001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000008E02280600000A7201000070038C07000001048C08000001280700000A056F0100002B2A8A02280600000A7201000070038C07000001048C08000001280700000A6F0200002B2A7202280600000A7229000070038C07000001280A00000A6F0B00000A2A7602280600000A7229000070038C07000001280A00000A046F0C00000A2A3A0203280D00000A02280D0000062A00000013300400220100000100001102030428020000060A02067B17000004723F000070280E00000A020206280700000616FE01726B000070280E00000A0202280F00000A6F1000000A067B13000004361802280F00000A6F1000000A067B16000004FE0516FE012B01167289000070280E00000A067B13000004166A364B02067B140000042803000006067B13000004D70B02280600000A067B140000046F1100000A2D1702067B1400000407281200000A6F1300000A2C03166A0B02067B14000004072804000006120002280F00000A6F1400000A7D14000004120002280F00000A6F1000000A7D13000004020304062801000006021202FE15040000021202037D060000041202047D070000041202067B140000047D080000041202067B130000047D0900000408280300002B2A5E02281600000A6F1700000A037B15000004FE0516FE012A0000133003008F0000000200001102280D0000060202280F00000A6F1400000A28030000060A0206166AFE037209010070280E00000A0202280F00000A6F1400000A166A28040000060202280F00000A6F1400000A06281200000A0B02076F1300000A7227010070280E00000A021202FE1507000002120202280F00000A6F1400000A7D100000041202067D1100000408280400002B076F1300000A2A0013300500180100000300001102280D00000602030428020000060A02067B17000004723F000070280E00000A02020628070000067249010070280E00000A1200167D17000004020304062801000006067B130000042D3402030402281800000A067B12000004280B000006021201FE15060000021201037D0E0000041201047D0F00000407280500002B2A02280600000A067B120000046F1100000A2C1402067B12000004067B1300000428040000062B2402067B12000004067B13000004281200000A0C02086F1300000A7227010070280E00000A02030402281800000A067B14000004280B000006021203FE15050000021203037D0A0000041203047D0B0000041203067B140000047D0C0000041203067B130000047D0D00000409280600002B2A133005002C0100000400001102280D00000602280F00000A6F1400000A0A0202280600000A066F1100000A727D010070280E00000A047E1900000A281A00000A2D03042B01030B02281B00000A0E046F0700002B0C02087B18000004166AFE0372B9010070280E00000A0202281800000A020605280C000006281A00000A72FD010070280E00000A02060528020000060D02097B1700000416FE017259020070280E00000A1204FE15080000021204077D12000004120402281600000A6F1700000A087B19000004D77D150000041204087B180000047D160000041204177D1700000411040D020605092801000006021205FE15030000021205067D010000041205057D020000041205097B150000047D030000041205097B120000047D050000041205097B160000047D040000041105280800002B172A1330080045000000050000110203166A729B020070198D0F0000012516058C07000001A225170E048C07000001A22518048C08000001A2166A281D00000A0A02066F1300000A72B5020070280E00000A2A0000001330080050000000050000110203166A72EB020070178D0F0000012516048C08000001A2166A281D00000A0A02066F1300000A2C10066F1E00000A750700000114FE032B011672FB020070280E00000A066F1E00000AA5070000012A6E0202280F00000A6F1000000A166AFE017235030070280E00000A2A42534A4201000100000000000C00000076342E302E33303331390000000005006C000000B8040000237E0000240500004804000023537472696E6773000000006C0900006C03000023555300D80C0000100000002347554944000000E80C00008801000023426C6F62000000000000000200000157150200090A000000FA013300160000010000001200000009000000190000000D000000180000001E00000010000000050000000100000002000000070000000800000000005402010000000000060054014E03060093014E03060040013B030F006E0300000A008301B9030A00E203B9030A00A903B9030A001500B9030A001501B9030A001D04B9030600F10073020A007401B9030A002901B9030600CB0173020600F90373020A00CB00B9030A003502B9030A001A03B903000000001D00000000000100010001001000FB0000001900010001000A011000230200002D0001000E000A0110000E0200002D0006000E000A011000E50100002D000A000E000A011000FA0100002D000E000E000A011000D20100002D0010000E000A011000BD0200002D0012000E000A011000660200002D0018000E000600E703B70006004200F80006004602FC000600B500FC0006000A03B7000600E703B70006004200F8000600EF02B70006008400FC000600E703B70006004200F8000600E802B70006007D00FC000600E703B70006004200F8000600A802B70006002D04FC0006000A03B70006007D00FC000600E802B70006004602FC000600A700FC0006009602FF000600A700FC0006008D02FC005020000000008100BA02020101007420000000008600AB020C0104009720000000008600930015010600B4200000000081009D001B010700D220000000008618350341000900E420000000008600840022010A00122200000000810052002A010C002C22000000008600A0006C000D00C822000000008600880022010D00EC23000000008600620030010F0024250000000081007F023C0113007825000000008100110348011700D425000000008100E0000600190000000100F003000002004A0000000300C90200000100F003000002004A0000000100B10300000100B10300000200C300000001003A0100000100F003000002004A0000000100A00200000100F003000002004A0000000100A103000002009503000003004A0000000400260000000100F003000002004A00000003007A0200000400DA0200000100F003000002004A00090035030100110035030600190035030A0029003503060061003503060031000B0110007100D0031500690013041C006900090429007100D003300069000100360069000B003B00310035034100310034044F003100D40055008100B1015A006900D7035E0031000103640051007D036C008100F602700031003102750031003C0281008900DD025A003100890370003900D502B70039003B04BB0031002603C30091000004C80031004F02DF005100BB01EB0021002B0082012E000B0051012E0013005A012E001B00790141002B008201430023008201A1002B008201C1002B008201E1002B00820101012B00820141012B00820161012B00820181012B008201C1012B008201E1012B00820101022B008201470086009300A800DA00048000000000000000000000000000000000E2030000040000000000000000000000EF002B00000000000200000000000000000000000000B903000000000300020004000200050002000600020007000200080002000900020011002400130024002B007C002B008E002B009E002B00A3003900D0002B00D500000000000047657455496E7436340053657455496E7436340055496E74323536003C4D6F64756C653E00646174610053797374656D2E507269766174652E436F72654C696200546F6B656E496400746F6B656E496400456E64426C6F636B52656163686564004F6E4E6F6E46756E6769626C65546F6B656E526563656976656400486967686573744269640041756374696F6E456E6400476574526566756E6400536574526566756E64005374617274696E675072696365007374617274696E6750726963650062616C616E636500494D657373616765006765745F4D65737361676500456E737572654E6F7450617961626C650056616C756554797065004E465441756374696F6E53746F7265006765745F53746174650049536D617274436F6E74726163745374617465004950657273697374656E7453746174650073746174650044656275676761626C6541747472696275746500436F6D70696C6174696F6E52656C61786174696F6E7341747472696275746500496E646578417474726962757465004465706C6F794174747269627574650052756E74696D65436F6D7061746962696C697479417474726962757465006765745F56616C7565006765745F52657475726E56616C756500537472696E670042616C616E6365526566756E6465644C6F670041756374696F6E456E64537563636565644C6F670041756374696F6E456E644661696C65644C6F670048696768657374426964557064617465644C6F670041756374696F6E537461727465644C6F670049426C6F636B006765745F426C6F636B00456E64426C6F636B0043616C6C00536D617274436F6E74726163742E646C6C0041756374696F6E506172616D0053797374656D0066726F6D005472616E73666572546F6B656E004475726174696F6E004F6E41756374696F6E0061756374696F6E00546F0047657441756374696F6E496E666F0053657441756374696F6E496E666F0061756374696F6E496E666F005A65726F00746F006765745F4E756D6265720048696768657374426964646572006765745F53656E646572005472616E736665720053656C6C6572004765744F776E6572004953657269616C697A6572006765745F53657269616C697A6572002E63746F720053797374656D2E446961676E6F73746963730053797374656D2E52756E74696D652E436F6D70696C6572536572766963657300446562756767696E674D6F646573006765745F53756363657373006765745F416464726573730066726F6D41646472657373006F70657261746F7241646472657373006164647265737300537472617469732E536D617274436F6E74726163747300466F726D6174004973436F6E747261637400536D617274436F6E747261637400636F6E7472616374004F626A65637400546F537472756374004765745374727563740053657453747275637400495472616E73666572526573756C7400416D6F756E7400417373657274006F705F457175616C69747900000027410075006300740069006F006E0049006E0066006F003A007B0030007D003A007B0031007D00001552006500660075006E0064003A007B0030007D00002B410075006300740069006F006E0020006900730020006E006F007400200066006F0075006E0064002E00001D410075006300740069006F006E00200065006E006400650064002E00007F540068006500200061006D006F0075006E00740020006900730020006E006F007400200068006900670068006500720020007400680061006E0020006800690067006800650073007400200062006900640064006500720020006F00720020007300740061007200740069006E0067002000700072006900630065002E00001D41007300730065007200740020006600610069006C00650064002E0000215400720061006E00730066006500720020006600610069006C00650064002E000033410075006300740069006F006E0020006900730020006E006F007400200065006E0064006500640020007900650074002E00003B5400680065002000430061006C006C006500720020006900730020006E006F00740020006100200063006F006E00740072006100630074002E000043500072006900630065002000730068006F0075006C006400200062006500200068006900670068006500720020007400680061006E0020007A00650072006F002E00005B5400680065002000730074006F0072006500200063006F006E007400720061006300740020006900730020006E006F00740020006F0077006E006500720020006F0066002000740068006500200074006F006B0065006E002E000041540068006500200074006F006B0065006E00200069007300200061006C007200650061006400790020006F006E002000610075006300740069006F006E002E0000195400720061006E007300660065007200460072006F006D000035540068006500200074006F006B0065006E0020007400720061006E00730066006500720020006600610069006C00650064002E00000F4F0077006E00650072004F00660000394700650074004F0077006E006500720020006D006500740068006F0064002000630061006C006C0020006600610069006C00650064002E00003554006800650020006D006500740068006F00640020006900730020006E006F0074002000700061007900610062006C0065002E000000EB1C822BEDD7A44FA806084221D756D10004200101080320000105200101111104200012350600030E0E1C1C07300102010E1E00040A011120063001011E000E0500020E0E1C0420010B0E052002010E0B05200101122507070311200B111005200201020E04200012410320000B05200102111D0720021229111D0B03200002042000111D06300101011E00040A01111004200012450707030B1229111C040A01111C0A07041120111812291114040A011118040A0111140E0706111D111D112411201120110C0306111D07000202111D111D0420001249073001011E001D05040A011124040A01110C04070112290B20051229111D0B0E1D1C0B0320001C087CEC85D7BEA7798E0306112102060B02060209200301111D112111200820021120111D11210520010B111D06200201111D0B07200201111D11210520010211200B200402111D111D11211D050B200401111D1121111D111D082002111D111D11210801000800000000001E01000100540216577261704E6F6E457863657074696F6E5468726F77730108010002000000000004010000000000000000000000000000000010000000000000000000000000000000A43400000000000000000000BE340000002000000000000000000000000000000000000000000000B0340000000000000000000000005F436F72446C6C4D61696E006D73636F7265652E646C6C0000000000FF2500200010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000003000000C000000D03400000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
```


Testnet/NFTAuctionStore/NFTAuctionStore/NFTAuctionStore.cs
```diff
@@ -0,0 +1,199 @@
...
+        public Address Bidder;
+        public ulong Bid;
+    }
+    public struct AuctionEndedLog
+    {
+        public Address Contract;
+        public ulong TokenId;
+        public Address HighestBidder;
+        public ulong HighestBid;
+    }
```
mrtpain, line 188: Indexes

Testnet/NFTAuctionStore/NFTAuctionStore/NFTAuctionStore.cs
```diff
@@ -0,0 +1,199 @@
...
+
+    private void EnsureNotPayable() => Assert(Message.Value == 0, "The method is not payable.");
+
+    public struct AuctionStartedLog
+    {
+        public Address Contract;
+        public ulong TokenId;
+        public ulong EndBlock;
+        public ulong startingPrice;
```
mrtpain, line 171: missing capitol S

Testnet/NFTAuctionStore/NFTAuctionStore/NFTAuctionStore.cs
```diff
@@ -0,0 +1,199 @@
...
+
+        Assert(result.Success && result.ReturnValue is bool success && success, "The token transfer failed. Be sure sender is approved to transfer token.");
+    }
+
+    private void SafeTransferToken(Address contract, ulong tokenId, Address from, Address to)
+    {
+        var result = Call(contract, 0, "SafeTransferFrom", new object[] { from, to, tokenId });
+
+        Assert(result.Success && result.ReturnValue is bool success && success, "The token transfer failed.");
+    }
```
mrtpain, line 148: Is there a difference between these two methods? Obviously method name is different but from a reviewers perspective, I have no idea.

Testnet/NFTAuctionStore/NFTAuctionStore/NFTAuctionStore.cs
```diff
@@ -0,0 +1,199 @@
...
+
+    private Address GetOwner(Address contract, ulong tokenId)
+    {
+        var result = Call(contract, 0, "GetOwner", new object[] { tokenId });
+
+        Assert(result.Success && result.ReturnValue is Address, "GetOwner method call failed.");
+
+        return (Address)result.ReturnValue;
```
mrtpain, line 156: Same as other PR, not familiar with the contract being called, but might be beneficial to enforce result != Address.Zero

Testnet/NFTAuctionStore/NFTAuctionStore/NFTAuctionStore.cs
```diff
@@ -0,0 +1,199 @@
...
+
+        Assert(amount > 0);
+
+        SetRefund(Message.Sender, 0);
+
+        var transfer = Transfer(this.Message.Sender, amount);
+
+        if (!transfer.Success)
+            SetRefund(Message.Sender, amount);
+
```
mrtpain, line 94: Will this really ever fail? 

Really, this would need to check `refundAmount <= Balance` to check the balance of the contract.

Ultimately, I don't see a need for this public method at all. If a someone bids higher than my previous bid, they should also have to pay the gas to refund my previous bid.

Testnet/NFTAuctionStore/NFTAuctionStore/NFTAuctionStore.cs
```diff
@@ -0,0 +1,199 @@
...
+        Assert(Message.Value > auction.HighestBid && Message.Value >= auction.StartingPrice, "The amount is not higher than highest bidder or starting price.");
+
+        if (auction.HighestBid > 0)
+        {
+            //refund for previous bidder
+            var balance = GetRefund(auction.HighestBidder);
+
+            SetRefund(auction.HighestBidder, balance + auction.HighestBid);
+        }
```
mrtpain, line 70: Would be cleaner if this just refunded the previous highest bidder immediately.

Testnet/NFTAuctionStore/NFTAuctionStore/NFTAuctionStore.cs
```diff
@@ -0,0 +1,199 @@
...
+
+    public void AuctionEnd(Address contract, ulong tokenId)
+    {
+        EnsureNotPayable();
+
+        var auction = GetAuctionInfo(contract, tokenId);
+
+        Assert(Block.Number >= auction.EndBlock, "Auction is not ended yet.");
+
+        Assert(!auction.Ended,"Auction end already executed.");
```
mrtpain, line 106: nit - space

Testnet/NFTAuctionStore/NFTAuctionStore/NFTAuctionStore.cs
```diff
@@ -0,0 +1,199 @@
...
+            var result = Transfer(auction.Seller, auction.HighestBid);
+
+            Assert(result.Success, "Transfer failed.");
+
+            SafeTransferToken(contract, tokenId, Address, auction.HighestBidder);
+        }
+        else
+        {
+            SafeTransferToken(contract, tokenId, Address, auction.Seller);
+        }
```
mrtpain, line 122: Not entirely accurate, if the starting price was higher than 0, the seller would never get their token back.

mrtpain (commented):
Nice work, couple comments/suggestions

Testnet/NFTAuctionStore/NFTAuctionStore/NFTAuctionStore.cs
YakupIpek: SafeTransferFrom works same as TransferFrom on NFT contract but in case of To is another smart contract then SafeTransferFrom does a call OnReceived method as notification on To(contract)

Testnet/NFTAuctionStore/NFTAuctionStore/NFTAuctionStore.cs
YakupIpek: If previous highest bidder is a contract and we send back via Transfer method, the contract can reject incoming transfer in Receive method. So in this case, nobody would be able to Bid for the token. 

This is reason why withdraw pattern is preferred.

Testnet/NFTAuctionStore/NFTAuctionStore/NFTAuctionStore.cs
YakupIpek: This is just safe by default approach. Caller can be another contract and may reject transfer. So it is there for worst case scenario.

YakupIpek:
@rowandh This is ready to merge.

Testnet/NFTAuctionStore/NFTAuctionStore/NFTAuctionStore.cs
YakupIpek: No need aditional checks because we compare value with current owner.

Testnet/NFTAuctionStore/NFTAuctionStore/NFTAuctionStore.cs
YakupIpek: > Ultimately, I don't see a need for this public method at all. If a someone bids higher than my previous bid, they should also have to pay the gas to refund my previous bid.

I am going to propose another pr for this case.

YakupIpek:
@mrtpain @rowandh I added aditional commits. Can you please review it asap ?

Testnet/NFTAuctionStore/NFTAuctionStore/NFTAuctionStore.cs
```diff
@@ -0,0 +1,228 @@
+﻿using Stratis.SmartContracts;
+using System;
+
+[Deploy]
+public class NFTAuctionStore : SmartContract //,INonFungibleTokenReceiver
```
mrtpain, line 5: Think the interface should be uncommented or removed.

mrtpain:
This lgtm nice work

bytecode and hash compiled locally match but depending on how the comment I mentioned above is handled, we may need to revalidate the hash and bytecode. 
```
96bebea40adcfe03d4ecab5ebb8e595213f1571f34228d9c00554c56a4d5729f
```

```
4D5A90000300000004000000FFFF0000B800000000000000400000000000000000000000000000000000000000000000000000000000000000000000800000000E1FBA0E00B409CD21B8014CCD21546869732070726F6772616D2063616E6E6F742062652072756E20696E20444F53206D6F64652E0D0D0A2400000000000000504500004C010200E9C903AB0000000000000000E00022200B0130000016000000020000000000007E3400000020000000400000000000100020000000020000040000000000000004000000000000000060000000020000000000000300408500001000001000000000100000100000000000001000000000000000000000002C3400004F000000000000000000000000000000000000000000000000000000004000000C000000103400001C0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000200000080000000000000000000000082000004800000000000000000000002E7465787400000084140000002000000016000000020000000000000000000000000000200000602E72656C6F6300000C00000000400000000200000018000000000000000000000000000040000042000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000060340000000000004800000002000500A0250000700E000001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000008E02280600000A7201000070038C07000001048C08000001280700000A056F0100002B2A8A02280600000A7201000070038C07000001048C08000001280700000A6F0200002B2A7202280600000A7229000070038C07000001280A00000A6F0B00000A2A7602280600000A7229000070038C07000001280A00000A046F0C00000A2A3A0203280D00000A02280D0000062A00000013300400F80000000100001102030428020000060A02067B17000004723F000070280E00000A020206280700000616FE01726B000070280E00000A0202280F00000A6F1000000A067B13000004361802280F00000A6F1000000A067B16000004FE0516FE012B01167289000070280E00000A067B13000004166A362102067B1400000428030000060B02067B1400000407067B13000004D72804000006120002280F00000A6F1100000A7D14000004120002280F00000A6F1000000A7D13000004020304062801000006021202FE15040000021202037D060000041202047D070000041202067B140000047D080000041202067B130000047D0900000408280300002B2A5E02281300000A6F1400000A037B15000004FE0516FE012A133003008F0000000200001102280D0000060202280F00000A6F1100000A28030000060A0206166AFE037209010070280E00000A0202280F00000A6F1100000A166A28040000060202280F00000A6F1100000A06281500000A0B02076F1600000A7227010070280E00000A021202FE1507000002120202280F00000A6F1100000A7D100000041202067D1100000408280400002B076F1600000A2A0013300500F10000000300001102280D00000602030428020000060A02067B17000004723F000070280E00000A02020628070000067249010070280E00000A1200167D17000004020304062801000006067B130000042D3402030402281700000A067B12000004280B000006021202FE15060000021202037D0E0000041202047D0F00000408280500002B2A02067B12000004067B13000004281500000A0B02076F1600000A7227010070280E00000A02030402281700000A067B14000004280B000006021203FE15050000021203037D0A0000041203047D0B0000041203067B140000047D0C0000041203067B130000047D0D00000409280600002B2A000000133005002C0100000400001102280D00000602280F00000A6F1100000A0A0202280600000A066F1800000A727D010070280E00000A047E1900000A281A00000A2D03042B01030B02281B00000A0E046F0700002B0C02087B18000004166AFE0372B9010070280E00000A0202281700000A020605280C000006281A00000A72FD010070280E00000A02060528020000060D02097B1700000416FE017259020070280E00000A1204FE15080000021204077D12000004120402281300000A6F1400000A087B19000004D77D150000041204087B180000047D160000041204177D1700000411040D020605092801000006021205FE15030000021205067D010000041205057D020000041205097B150000047D030000041205097B120000047D050000041205097B160000047D040000041105280800002B172A1330080045000000050000110203166A729B020070198D0F0000012516058C07000001A225170E048C07000001A22518048C08000001A2166A281D00000A0A02066F1600000A72B5020070280E00000A2A0000001330080050000000050000110203166A72EB020070178D0F0000012516048C08000001A2166A281D00000A0A02066F1600000A2C10066F1E00000A750700000114FE032B011672FB020070280E00000A066F1E00000AA5070000012A6E0202280F00000A6F1000000A166AFE017235030070280E00000A2A42534A4201000100000000000C00000076342E302E33303331390000000005006C000000B8040000237E0000240500004804000023537472696E6773000000006C0900006C03000023555300D80C0000100000002347554944000000E80C00008801000023426C6F62000000000000000200000157150200090A000000FA013300160000010000001200000009000000190000000D000000180000001E00000010000000050000000100000002000000070000000800000000005402010000000000060054014E03060093014E03060040013B030F006E0300000A008301B9030A00E203B9030A00A903B9030A001500B9030A001501B9030A001D04B9030600F10073020A007401B9030A002901B9030600CB0173020600F90373020A00CB00B9030A003502B9030A001A03B903000000001D00000000000100010001001000FB0000001900010001000A011000230200002D0001000E000A0110000E0200002D0006000E000A011000E50100002D000A000E000A011000FA0100002D000E000E000A011000D20100002D0010000E000A011000BD0200002D0012000E000A011000660200002D0018000E000600E703B70006004200F80006004602FC000600B500FC0006000A03B7000600E703B70006004200F8000600EF02B70006008400FC000600E703B70006004200F8000600E802B70006007D00FC000600E703B70006004200F8000600A802B70006002D04FC0006000A03B70006007D00FC000600E802B70006004602FC000600A700FC0006009602FF000600A700FC0006008D02FC005020000000008100BA02020101007420000000008600AB020C0104009720000000008600930015010600B4200000000081009D001B010700D220000000008618350341000900E420000000008600840022010A00E82100000000810052002A010C000022000000008600A00084000D009C22000000008600880022010D009C23000000008600620030010F00D4240000000081007F023C01130028250000000081001103480117008425000000008100E0000600190000000100F003000002004A0000000300C90200000100F003000002004A0000000100B10300000100B10300000200C300000001003A0100000100F003000002004A0000000100A00200000100F003000002004A0000000100A103000002009503000003004A0000000400260000000100F003000002004A00000003007A0200000400DA0200000100F003000002004A00090035030100110035030600190035030A0029003503060061003503060031000B0110007100D0031500690013041C006900090429007100D003300069000100360069000B003B00310035034100310034044F003100D40055008100B1015A008100F6025E0031003102630031003C026F008900DD025A00310001037C0051007D038400310089035E006900D703B1003900D502B70039003B04BB0031002603C30091000004C80031004F02DF005100BB01EB0021002B0082012E000B0051012E0013005A012E001B00790141002B008201430023008201A1002B008201C1002B008201E1002B00820101012B00820141012B00820161012B00820181012B008201C1012B008201E1012B00820101022B008201470074008D00A200DA00048000000000000000000000000000000000E2030000040000000000000000000000EF002B00000000000200000000000000000000000000B9030000000003000200040002000500020006000200070002000800020009000200110024001300240025006A00250088002500980025009D003900D0002500D500000000000047657455496E7436340053657455496E7436340055496E74323536003C4D6F64756C653E00646174610053797374656D2E507269766174652E436F72654C696200546F6B656E496400746F6B656E496400456E64426C6F636B52656163686564004F6E4E6F6E46756E6769626C65546F6B656E526563656976656400486967686573744269640041756374696F6E456E6400476574526566756E6400536574526566756E64005374617274696E675072696365007374617274696E6750726963650062616C616E636500494D657373616765006765745F4D65737361676500456E737572654E6F7450617961626C650056616C756554797065004E465441756374696F6E53746F7265006765745F53746174650049536D617274436F6E74726163745374617465004950657273697374656E7453746174650073746174650044656275676761626C6541747472696275746500436F6D70696C6174696F6E52656C61786174696F6E7341747472696275746500496E646578417474726962757465004465706C6F794174747269627574650052756E74696D65436F6D7061746962696C697479417474726962757465006765745F56616C7565006765745F52657475726E56616C756500537472696E670042616C616E6365526566756E6465644C6F670041756374696F6E456E64537563636565644C6F670041756374696F6E456E644661696C65644C6F670048696768657374426964557064617465644C6F670041756374696F6E537461727465644C6F670049426C6F636B006765745F426C6F636B00456E64426C6F636B0043616C6C00536D617274436F6E74726163742E646C6C0041756374696F6E506172616D0053797374656D0066726F6D005472616E73666572546F6B656E004475726174696F6E004F6E41756374696F6E0061756374696F6E00546F0047657441756374696F6E496E666F0053657441756374696F6E496E666F0061756374696F6E496E666F005A65726F00746F006765745F4E756D6265720048696768657374426964646572006765745F53656E646572005472616E736665720053656C6C6572004765744F776E6572004953657269616C697A6572006765745F53657269616C697A6572002E63746F720053797374656D2E446961676E6F73746963730053797374656D2E52756E74696D652E436F6D70696C6572536572766963657300446562756767696E674D6F646573006765745F53756363657373006765745F416464726573730066726F6D41646472657373006F70657261746F7241646472657373006164647265737300537472617469732E536D617274436F6E74726163747300466F726D6174004973436F6E747261637400536D617274436F6E747261637400636F6E7472616374004F626A65637400546F537472756374004765745374727563740053657453747275637400495472616E73666572526573756C7400416D6F756E7400417373657274006F705F457175616C69747900000027410075006300740069006F006E0049006E0066006F003A007B0030007D003A007B0031007D00001552006500660075006E0064003A007B0030007D00002B410075006300740069006F006E0020006900730020006E006F007400200066006F0075006E0064002E00001D410075006300740069006F006E00200065006E006400650064002E00007F540068006500200061006D006F0075006E00740020006900730020006E006F007400200068006900670068006500720020007400680061006E0020006800690067006800650073007400200062006900640064006500720020006F00720020007300740061007200740069006E0067002000700072006900630065002E00001D41007300730065007200740020006600610069006C00650064002E0000215400720061006E00730066006500720020006600610069006C00650064002E000033410075006300740069006F006E0020006900730020006E006F007400200065006E0064006500640020007900650074002E00003B5400680065002000430061006C006C006500720020006900730020006E006F00740020006100200063006F006E00740072006100630074002E000043500072006900630065002000730068006F0075006C006400200062006500200068006900670068006500720020007400680061006E0020007A00650072006F002E00005B5400680065002000730074006F0072006500200063006F006E007400720061006300740020006900730020006E006F00740020006F0077006E006500720020006F0066002000740068006500200074006F006B0065006E002E000041540068006500200074006F006B0065006E00200069007300200061006C007200650061006400790020006F006E002000610075006300740069006F006E002E0000195400720061006E007300660065007200460072006F006D000035540068006500200074006F006B0065006E0020007400720061006E00730066006500720020006600610069006C00650064002E00000F4F0077006E00650072004F00660000394700650074004F0077006E006500720020006D006500740068006F0064002000630061006C006C0020006600610069006C00650064002E00003554006800650020006D006500740068006F00640020006900730020006E006F0074002000700061007900610062006C0065002E00000091E67815E998AB4CAB0AC54BDF46A3520004200101080320000105200101111104200012350600030E0E1C1C07300102010E1E00040A011120063001011E000E0500020E0E1C0420010B0E052002010E0B05200101122507070311200B111005200201020E04200012410320000B042000111D06300101011E00040A01111004200012450707030B1229111C0720021229111D0B03200002040A01111C0A07041120122911181114040A011118040A0111140E0706111D111D112411201120110C05200102111D0306111D07000202111D111D0420001249073001011E001D05040A011124040A01110C04070112290B20051229111D0B0E1D1C0B0320001C087CEC85D7BEA7798E0306112102060B02060209200301111D112111200820021120111D11210520010B111D06200201111D0B07200201111D11210520010211200B200402111D111D11211D050B200401111D1121111D111D082002111D111D11210801000800000000001E01000100540216577261704E6F6E457863657074696F6E5468726F777301080100020000000000040100000000000000000000000000000000100000000000000000000000000000005434000000000000000000006E34000000200000000000000000000000000000000000000000000060340000000000000000000000005F436F72446C6C4D61696E006D73636F7265652E646C6C0000000000FF25002000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000003000000C000000803400000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
```

Testnet/NFTAuctionStore/NFTAuctionStore/NFTAuctionStore.cs
```diff
@@ -0,0 +1,228 @@
...
+
+        if (auction.HighestBid == 0)
+        {
+            TransferToken(contract, tokenId, Address, auction.Seller);
+
+            Log(new AuctionEndFailedLog { Contract = contract, TokenId = tokenId });
+            return;
+        }
+
+        var result = Transfer(auction.Seller, auction.HighestBid);
```
rowandh, line 103: These funds should instead be a withdrawable balance for the seller.

rowandh:
The implementation seems alright, though IMO there are a few problems with this design which we have discussed elsewhere, so I will just outline them here for posterity's sake:
* It would be preferable to use a new contract per auction. This is a simpler design, avoids having higher balances in a single contract, avoids an issue with one auction potentially influencing another, and allows new auctions to easily use a different version of the auction contract if necessary, for example due to an upgrade or just a custom auction requirement.
* It would be much simpler if a bid was just the sum of all funds sent to the contract buy a user. The highest bidder would be the user who sent the most funds to the contract, and their funds would be locked. All other bidders could withdraw their funds at any time.

This would also avoid these issues with the current design:
* If a user wants to bid again, they need to first withdraw their funds.
* If a user wants to increase their bid, they must do so before they can withdraw their old bid, meaning they need more than twice the amount of funds than their total bid.

Testnet/NFTAuctionStore/NFTAuctionStore/NFTAuctionStore.cs
YakupIpek: It is not possible to uncomment it. ATM it is a kind of documentation.

Testnet/NFTAuctionStore/NFTAuctionStore/NFTAuctionStore.cs
YakupIpek: I made improvement in here. I refund when seller is a contract

Testnet/NFTAuctionStore/NFTAuctionStore/NFTAuctionStore.cs
mrtpain: You can inline the interface with this contract to compile with a single file, or compile the entire project directory if it's in a different file. But either way, leaving as comment for documentation is fine too 👍

mrtpain (approved):
Re-validated the contracts with the latest commits. Changes lgtm 

**Hash**
```
bc12bedefb91d22cdd7318950cb1971825a26d2fa6afec04e56b18f11898b4c7
```

**Bytecode**
```
4D5A90000300000004000000FFFF0000B800000000000000400000000000000000000000000000000000000000000000000000000000000000000000800000000E1FBA0E00B409CD21B8014CCD21546869732070726F6772616D2063616E6E6F742062652072756E20696E20444F53206D6F64652E0D0D0A2400000000000000504500004C010200D23F989D0000000000000000E00022200B013000001600000002000000000000CE3400000020000000400000000000100020000000020000040000000000000004000000000000000060000000020000000000000300408500001000001000000000100000100000000000001000000000000000000000007C3400004F000000000000000000000000000000000000000000000000000000004000000C000000603400001C0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000200000080000000000000000000000082000004800000000000000000000002E74657874000000D4140000002000000016000000020000000000000000000000000000200000602E72656C6F6300000C000000004000000002000000180000000000000000000000000000400000420000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000B0340000000000004800000002000500F0250000700E000001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000008E02280600000A7201000070038C07000001048C08000001280700000A056F0100002B2A8A02280600000A7201000070038C07000001048C08000001280700000A6F0200002B2A7202280600000A7229000070038C07000001280A00000A6F0B00000A2A7602280600000A7229000070038C07000001280A00000A046F0C00000A2A3A0203280D00000A02280D0000062A00000013300400220100000100001102030428020000060A02067B17000004723F000070280E00000A020206280700000616FE01726B000070280E00000A0202280F00000A6F1000000A067B13000004361802280F00000A6F1000000A067B16000004FE0516FE012B01167289000070280E00000A067B13000004166A364B02067B140000042803000006067B13000004D70B02280600000A067B140000046F1100000A2D1702067B1400000407281200000A6F1300000A2C03166A0B02067B14000004072804000006120002280F00000A6F1400000A7D14000004120002280F00000A6F1000000A7D13000004020304062801000006021202FE15040000021202037D060000041202047D070000041202067B140000047D080000041202067B130000047D0900000408280300002B2A5E02281600000A6F1700000A037B15000004FE0516FE012A0000133003008F0000000200001102280D0000060202280F00000A6F1400000A28030000060A0206166AFE037209010070280E00000A0202280F00000A6F1400000A166A28040000060202280F00000A6F1400000A06281200000A0B02076F1300000A7227010070280E00000A021202FE1507000002120202280F00000A6F1400000A7D100000041202067D1100000408280400002B076F1300000A2A0013300500180100000300001102280D00000602030428020000060A02067B17000004723F000070280E00000A02020628070000067249010070280E00000A1200167D17000004020304062801000006067B130000042D3402030402281800000A067B12000004280B000006021201FE15060000021201037D0E0000041201047D0F00000407280500002B2A02280600000A067B120000046F1100000A2C1402067B12000004067B1300000428040000062B2402067B12000004067B13000004281200000A0C02086F1300000A7227010070280E00000A02030402281800000A067B14000004280B000006021203FE15050000021203037D0A0000041203047D0B0000041203067B140000047D0C0000041203067B130000047D0D00000409280600002B2A133005002C0100000400001102280D00000602280F00000A6F1400000A0A0202280600000A066F1100000A727D010070280E00000A047E1900000A281A00000A2D03042B01030B02281B00000A0E046F0700002B0C02087B18000004166AFE0372B9010070280E00000A0202281800000A020605280C000006281A00000A72FD010070280E00000A02060528020000060D02097B1700000416FE017259020070280E00000A1204FE15080000021204077D12000004120402281600000A6F1700000A087B19000004D77D150000041204087B180000047D160000041204177D1700000411040D020605092801000006021205FE15030000021205067D010000041205057D020000041205097B150000047D030000041205097B120000047D050000041205097B160000047D040000041105280800002B172A1330080045000000050000110203166A729B020070198D0F0000012516058C07000001A225170E048C07000001A22518048C08000001A2166A281D00000A0A02066F1300000A72B5020070280E00000A2A0000001330080050000000050000110203166A72EB020070178D0F0000012516048C08000001A2166A281D00000A0A02066F1300000A2C10066F1E00000A750700000114FE032B011672FB020070280E00000A066F1E00000AA5070000012A6E0202280F00000A6F1000000A166AFE017235030070280E00000A2A42534A4201000100000000000C00000076342E302E33303331390000000005006C000000B8040000237E0000240500004804000023537472696E6773000000006C0900006C03000023555300D80C0000100000002347554944000000E80C00008801000023426C6F62000000000000000200000157150200090A000000FA013300160000010000001200000009000000190000000D000000180000001E00000010000000050000000100000002000000070000000800000000005402010000000000060054014E03060093014E03060040013B030F006E0300000A008301B9030A00E203B9030A00A903B9030A001500B9030A001501B9030A001D04B9030600F10073020A007401B9030A002901B9030600CB0173020600F90373020A00CB00B9030A003502B9030A001A03B903000000001D00000000000100010001001000FB0000001900010001000A011000230200002D0001000E000A0110000E0200002D0006000E000A011000E50100002D000A000E000A011000FA0100002D000E000E000A011000D20100002D0010000E000A011000BD0200002D0012000E000A011000660200002D0018000E000600E703B70006004200F80006004602FC000600B500FC0006000A03B7000600E703B70006004200F8000600EF02B70006008400FC000600E703B70006004200F8000600E802B70006007D00FC000600E703B70006004200F8000600A802B70006002D04FC0006000A03B70006007D00FC000600E802B70006004602FC000600A700FC0006009602FF000600A700FC0006008D02FC005020000000008100BA02020101007420000000008600AB020C0104009720000000008600930015010600B4200000000081009D001B010700D220000000008618350341000900E420000000008600840022010A00122200000000810052002A010C002C22000000008600A0006C000D00C822000000008600880022010D00EC23000000008600620030010F0024250000000081007F023C0113007825000000008100110348011700D425000000008100E0000600190000000100F003000002004A0000000300C90200000100F003000002004A0000000100B10300000100B10300000200C300000001003A0100000100F003000002004A0000000100A00200000100F003000002004A0000000100A103000002009503000003004A0000000400260000000100F003000002004A00000003007A0200000400DA0200000100F003000002004A00090035030100110035030600190035030A0029003503060061003503060031000B0110007100D0031500690013041C006900090429007100D003300069000100360069000B003B00310035034100310034044F003100D40055008100B1015A006900D7035E0031000103640051007D036C008100F602700031003102750031003C0281008900DD025A003100890370003900D502B70039003B04BB0031002603C30091000004C80031004F02DF005100BB01EB0021002B0082012E000B0051012E0013005A012E001B00790141002B008201430023008201A1002B008201C1002B008201E1002B00820101012B00820141012B00820161012B00820181012B008201C1012B008201E1012B00820101022B008201470086009300A800DA00048000000000000000000000000000000000E2030000040000000000000000000000EF002B00000000000200000000000000000000000000B903000000000300020004000200050002000600020007000200080002000900020011002400130024002B007C002B008E002B009E002B00A3003900D0002B00D500000000000047657455496E7436340053657455496E7436340055496E74323536003C4D6F64756C653E00646174610053797374656D2E507269766174652E436F72654C696200546F6B656E496400746F6B656E496400456E64426C6F636B52656163686564004F6E4E6F6E46756E6769626C65546F6B656E526563656976656400486967686573744269640041756374696F6E456E6400476574526566756E6400536574526566756E64005374617274696E675072696365007374617274696E6750726963650062616C616E636500494D657373616765006765745F4D65737361676500456E737572654E6F7450617961626C650056616C756554797065004E465441756374696F6E53746F7265006765745F53746174650049536D617274436F6E74726163745374617465004950657273697374656E7453746174650073746174650044656275676761626C6541747472696275746500436F6D70696C6174696F6E52656C61786174696F6E7341747472696275746500496E646578417474726962757465004465706C6F794174747269627574650052756E74696D65436F6D7061746962696C697479417474726962757465006765745F56616C7565006765745F52657475726E56616C756500537472696E670042616C616E6365526566756E6465644C6F670041756374696F6E456E64537563636565644C6F670041756374696F6E456E644661696C65644C6F670048696768657374426964557064617465644C6F670041756374696F6E537461727465644C6F670049426C6F636B006765745F426C6F636B00456E64426C6F636B0043616C6C00536D617274436F6E74726163742E646C6C0041756374696F6E506172616D0053797374656D0066726F6D005472616E73666572546F6B656E004475726174696F6E004F6E41756374696F6E0061756374696F6E00546F0047657441756374696F6E496E666F0053657441756374696F6E496E666F0061756374696F6E496E666F005A65726F00746F006765745F4E756D6265720048696768657374426964646572006765745F53656E646572005472616E736665720053656C6C6572004765744F776E6572004953657269616C697A6572006765745F53657269616C697A6572002E63746F720053797374656D2E446961676E6F73746963730053797374656D2E52756E74696D652E436F6D70696C6572536572766963657300446562756767696E674D6F646573006765745F53756363657373006765745F416464726573730066726F6D41646472657373006F70657261746F7241646472657373006164647265737300537472617469732E536D617274436F6E74726163747300466F726D6174004973436F6E747261637400536D617274436F6E747261637400636F6E7472616374004F626A65637400546F537472756374004765745374727563740053657453747275637400495472616E73666572526573756C7400416D6F756E7400417373657274006F705F457175616C69747900000027410075006300740069006F006E0049006E0066006F003A007B0030007D003A007B0031007D00001552006500660075006E0064003A007B0030007D00002B410075006300740069006F006E0020006900730020006E006F007400200066006F0075006E0064002E00001D410075006300740069006F006E00200065006E006400650064002E00007F540068006500200061006D006F0075006E00740020006900730020006E006F007400200068006900670068006500720020007400680061006E0020006800690067006800650073007400200062006900640064006500720020006F00720020007300740061007200740069006E0067002000700072006900630065002E00001D41007300730065007200740020006600610069006C00650064002E0000215400720061006E00730066006500720020006600610069006C00650064002E000033410075006300740069006F006E0020006900730020006E006F007400200065006E0064006500640020007900650074002E00003B5400680065002000430061006C006C006500720020006900730020006E006F00740020006100200063006F006E00740072006100630074002E000043500072006900630065002000730068006F0075006C006400200062006500200068006900670068006500720020007400680061006E0020007A00650072006F002E00005B5400680065002000730074006F0072006500200063006F006E007400720061006300740020006900730020006E006F00740020006F0077006E006500720020006F0066002000740068006500200074006F006B0065006E002E000041540068006500200074006F006B0065006E00200069007300200061006C007200650061006400790020006F006E002000610075006300740069006F006E002E0000195400720061006E007300660065007200460072006F006D000035540068006500200074006F006B0065006E0020007400720061006E00730066006500720020006600610069006C00650064002E00000F4F0077006E00650072004F00660000394700650074004F0077006E006500720020006D006500740068006F0064002000630061006C006C0020006600610069006C00650064002E00003554006800650020006D006500740068006F00640020006900730020006E006F0074002000700061007900610062006C0065002E000000EB1C822BEDD7A44FA806084221D756D10004200101080320000105200101111104200012350600030E0E1C1C07300102010E1E00040A011120063001011E000E0500020E0E1C0420010B0E052002010E0B05200101122507070311200B111005200201020E04200012410320000B05200102111D0720021229111D0B03200002042000111D06300101011E00040A01111004200012450707030B1229111C040A01111C0A07041120111812291114040A011118040A0111140E0706111D111D112411201120110C0306111D07000202111D111D0420001249073001011E001D05040A011124040A01110C04070112290B20051229111D0B0E1D1C0B0320001C087CEC85D7BEA7798E0306112102060B02060209200301111D112111200820021120111D11210520010B111D06200201111D0B07200201111D11210520010211200B200402111D111D11211D050B200401111D1121111D111D082002111D111D11210801000800000000001E01000100540216577261704E6F6E457863657074696F6E5468726F77730108010002000000000004010000000000000000000000000000000010000000000000000000000000000000A43400000000000000000000BE340000002000000000000000000000000000000000000000000000B0340000000000000000000000005F436F72446C6C4D61696E006D73636F7265652E646C6C0000000000FF2500200010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000003000000C000000D03400000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
```

PR #44, opened 2021-07-24
[Testnet]  Additional methods that support most basic use cases for Non Fungible Token

YakupIpek:
INonFungibleTokenMetadata interface implemented. 
OwnerOnlyMinting option added. 
Auto token id added for minting. 
Mint/SafeMint/Burn methods added. 
TransferOwnership method added.


Testnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
```diff
@@ -467,6 +489,103 @@ private void CanTransfer(ulong tokenId)
     /// <param name="tokenId">ID of the NFT to validate.</param>
     private void ValidNFToken(ulong tokenId)
     {
-        Assert(GetIdToOwner(tokenId) != Address.Zero);
+        Address tokenOwner = GetIdToOwner(tokenId);
+        EnsureAddressIsNotEmpty(tokenOwner);
+    }
+
+    /// <summary>
+    /// Sets the contract owner who can mint/bur
```
mrtpain, line 497: "burn" small typo in comment

Testnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
```diff
@@ -467,6 +489,103 @@ private void CanTransfer(ulong tokenId)
...
+    /// </summary>
+    /// <param name="owner"></param>
+    public void TransferOwnership(Address owner)
```
mrtpain, line 500: Small type in comment "burn" - missing owner comment

Testnet/NonFungibleToken/NonFungibleToken/INonFungibleTokenMetadata.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace NonFungibleTokenContract
+{
+    public interface INonFungibleTokenMetadata
+    {
+        string Name { get; }
+        string Symbol { get; }
+
+        string TokenURI(ulong tokenId);
+    }
+}
```
mrtpain, line 14: Is this really even used anywhere? 

Should the namespace be included? Should the interface be inline with the rest of the compiled contract code?

YakupIpek: These interfaces are a kind of documentation or standardization of current NFT contract. And it is not harmful to keep it in repository. Actually these interfaces are not included compiled bytecode. 

We should to move these interfaces into Stratis.SmartContracts.Standards library in future.

YakupIpek:
@rowandh @mrtpain Is there anything else ? I want to merge this pr and add some small updates as next pr.

noescape00:
Btw it was already whitelisted on testnet on Friday. 

[NFT.txt](https://github.com/stratisproject/CirrusSmartContracts/files/6955649/NFT.txt)

YakupIpek:
@noescape00 there is a bug related to Clear method on SC infrastructure so current implementation will not work as expected for some cases.

I will push workaround for it as next pr until that bug is fixed.

noescape00:
@YakupIpek maybe I'm missing smth, but let's say I received an NFT, I know smart contract address and that I have one. 
I want to burn my NFT and for that I need ID of my NFT, where do I get it from?

YakupIpek:
> @YakupIpek maybe I'm missing smth, but let's say I received an NFT, I know smart contract address and that I have one.
> I want to burn my NFT and for that I need ID of my NFT, where do I get it from?

Good question, If you know token id then you can verify it easily on smart contract that you are owner of that token or not. 

But you can not obtain transferred token id from contract directly. You need application or NFT Wallet that tracks logs and show you received tokens.

There is Enumerable interface of NFT contract that you can directly query and see which tokens you own. But still that interface doesn't tell you latest transferred token id. Here is interface. 

```
interface ERC721Enumerable /* is ERC721 */ {
    /// @notice Count NFTs tracked by this contract
    /// @return A count of valid NFTs tracked by this contract, where each one of
    ///  them has an assigned and queryable owner not equal to the zero address
    function totalSupply() external view returns (uint256);

    /// @notice Enumerate valid NFTs
    /// @dev Throws if `_index` >= `totalSupply()`.
    /// @param _index A counter less than `totalSupply()`
    /// @return The token identifier for the `_index`th NFT,
    ///  (sort order not specified)
    function tokenByIndex(uint256 _index) external view returns (uint256);

    /// @notice Enumerate NFTs assigned to an owner
    /// @dev Throws if `_index` >= `balanceOf(_owner)` or if
    ///  `_owner` is the zero address, representing invalid NFTs.
    /// @param _owner An address where we are interested in NFTs owned by them
    /// @param _index A counter less than `balanceOf(_owner)`
    /// @return The token identifier for the `_index`th NFT assigned to `_owner`,
    ///   (sort order not specified)
    function tokenOfOwnerByIndex(address _owner, uint256 _index) external view returns (uint256);
}
```
What you need is tokenOfOwnerByIndex method. 

That Interface implemented but i prune that interface because it is likely not that needed for most of contracts. Unfortunately inheritance is not supported atm so i can not provide inheritable one for extend current implementation. 

You can see that implementation in here. 
https://github.com/stratisproject/CirrusSmartContracts/pull/28/files#diff-e0dc938a01c87df6cbcc2d4d613baf339a5a93017dbb192520710806f6fa5640R231

noescape00:
`This interface bring additional storage costs.`

Network is empty, maybe we can afford it? 

Getting info on what user owns via logs sounds like a workaround

YakupIpek:
> `This interface bring additional storage costs.`
> 
> Network is empty, maybe we can afford it?
> 
> Getting info on what user owns via logs sounds like a workaround

It is not a workaround. Most of the applications doesn't need it even. 

I am thinking how i can provide 2 implementations or Enumerable methods as abstract class optionally. 

 @rowandh do you have advise on it ? I have to create another branch for that implementation and whitelist it ? 

Another alternative is the create Factory contract which can deploy contracts that has inheritance if you remember.

Testnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
```diff
@@ -419,7 +441,7 @@ private void LogTransfer(Address from, Address to, ulong tokenId)
     /// </summary>
     /// <param name="owner">The owner address.</param>
     /// <param name="operatorAddress">The approved address.</param>
-    /// <param name="tokenId">The NFT ID.</param>
+    /// <param name="tokenId">The NFT ID.</ >
```
rowandh, line 444:
```suggestion
    /// <param name="tokenId">The NFT ID.</param>
```

rowandh:
> > `This interface bring additional storage costs.`
> > Network is empty, maybe we can afford it?
> > Getting info on what user owns via logs sounds like a workaround
> 
> It is not a workaround. Most of the applications doesn't need it even.
> 
> I am thinking how i can provide 2 implementations or Enumerable methods as abstract class optionally.
> 
> @rowandh do you have advise on it ? I have to create another branch for that implementation and whitelist it ?
> 
> Another alternative is the create Factory contract which can deploy contracts that has inheritance if you remember.

I wouldn't worry about the complexity of having 2 contracts and a factory at the moment, I think it's fine to include the enumerable methods here if they are useful.

noescape00:
Can we have public method like GetOwnerToIds that will take in address and return an array of Ids that this address owns?

noescape00:
Also TokenIdCounter probably should initialized at 0. 
I've created a new NFT, minted 6 times, tried to transfer with id 0,1,2,3, all failed, so I assume TokenIdCounter initialized at some random value

noescape00:
Also it would be awesome if I could transfer a token without knowing it's id.
NFTs of same type are fungible so I dont see why I would need to transfer specific token.

YakupIpek:
@noescape00  I will try to provide Enumerable interface implementation. 

> Also TokenIdCounter probably should initialized at 0.
> I've created a new NFT, minted 6 times, tried to transfer with id 0,1,2,3, all failed, so I assume TokenIdCounter initialized at some random value.

Actually it is better to return tokenId from mint methods. I will do this. 

TokenIdCounter starts from 1 and incremented by 1 always. so 1,2,3 methods should to be fine to transfer if sender address is trying to transfer them.

YakupIpek:
> Also it would be awesome if I could transfer a token without knowing it's id.
> NFTs of same type are fungible so I dont see why I would need to transfer specific token.

NFTs are not fungible. They are individual assets. You have to decide which asset you try to transfer so you need to know tokenId.

noescape00:
`Actually it is better to return tokenId from mint methods. I will do this.`

That will solve case when minter sends token, but when someone else sends one they basically nave no way to do so.

noescape00:
`NFTs are not fungible. They are individual assets. You have to decide which asset you try to transfer so you need to know tokenId.`

Some are. 

So let's say NFT is a sword and it has 2 parameters: attack speed and attack damage. 
2 nfts with those parameters equal are fungible. 

But ok, scrap the ability to transfer without knowing the ID, just the ability to get ID will solve the problem.

YakupIpek:
@noescape00 Next pr is going to return token id from mint methods. We need to merge this asap. @rowandh

noescape00:
Returning token id in mint doesn't fix the problem when user got nft and wants to transfer it because id is unknown. 

Collecting and storing logs from all contracts is an insufficient overkill and is not implemented as far as I know, which basically makes transferring nfts not feasible.

YakupIpek:
Collecting logs is off-chain process and it is possible. This is how nft marketplace app works. Also Marketplace app doesn't need those Enumerable methods as well. 

Even that interface will not tell which token is transferred to you but it will you which tokens you have atm. I can help you better if i know more about your use case and how expect to do it. If a smart contracts wants to notified that any token is transferred then smart contract should to implement INonFungibleTokenReceiver but NFT contract will only notify destination contract if transfer is happened from SafeTransfer methods but not from Transfer.... methods. 

```
    /// <summary>
    /// Interface for a non-fungible token receiver.
    /// </summary>
    public interface INonFungibleTokenReceiver
    {
        /// <summary>
        /// Handle the receipt of a NFT. The smart contract calls this function on the
        /// recipient after a transfer. This function MAY throw or return false to revert and reject the transfer.
        /// Return true if the transfer is ok.
        /// </summary>
        /// <param name="operatorAddress">The address which called safeTransferFrom function.</param>
        /// <param name="fromAddress">The address which previously owned the token.</param>
        /// <param name="tokenId">The NFT identifier which is being transferred.</param>
        /// <param name="data">Additional data with no specified format.</param>
        /// <returns>A bool indicating the resulting operation.</returns>
        bool OnNonFungibleTokenReceived(Address operatorAddress, Address fromAddress, ulong tokenId, byte[] data);
    }

```
I will implement this interface as next pr. This may preference for some applications for some reasons. 

You should to call GetBalance to see how many tokens you have then you can enumerate it by calling it multiple times by incrementing index parameter until reach balance. 

function tokenOfOwnerByIndex(address _owner, uint256 _index) external view returns (uint256);


interface ERC721Enumerable /* is ERC721 */ {
    /// @notice Count NFTs tracked by this contract
    /// @return A count of valid NFTs tracked by this contract, where each one of
    ///  them has an assigned and queryable owner not equal to the zero address
    function totalSupply() external view returns (uint256);

    /// @notice Enumerate valid NFTs
    /// @dev Throws if `_index` >= `totalSupply()`.
    /// @param _index A counter less than `totalSupply()`
    /// @return The token identifier for the `_index`th NFT,
    ///  (sort order not specified)
    function tokenByIndex(uint256 _index) external view returns (uint256);

    /// @notice Enumerate NFTs assigned to an owner
    /// @dev Throws if `_index` >= `balanceOf(_owner)` or if
    ///  `_owner` is the zero address, representing invalid NFTs.
    /// @param _owner An address where we are interested in NFTs owned by them
    /// @param _index A counter less than `balanceOf(_owner)`
    /// @return The token identifier for the `_index`th NFT assigned to `_owner`,
    ///   (sort order not specified)
    function tokenOfOwnerByIndex(address _owner, uint256 _index) external view returns (uint256);
}

mrtpain:
> Can we have public method like GetOwnerToIds that will take in address and return an array of Ids that this address owns?

We can't safely do this because it would have to have _some_ limit of how many NFTs an address could own or gas limits could prevent the user from retrieving the values if they hold too many NFTs.

mrtpain:
@noescape00 @YakupIpek can the token Id just be logged in the transactions logs somewhere on transfer? Then the receivers just need to inspect the tx Logs of the transfer to see the Id. It's not ideal but there's not really another safe way unless you do some time of pagination to get all owned tokens which can still run into gas cost issues. 

There is also a FN related issue that will come into play currently if a list of anything is returned, event uint TokenIds. FN `return` value won't output array return values of any type currently.

YakupIpek:
> > Can we have public method like GetOwnerToIds that will take in address and return an array of Ids that this address owns?
> 
> We can't safely do this because it would have to have _some_ limit of how many NFTs an address could own or gas limits could prevent the user from retrieving the values if they hold too many NFTs.

Most likely that accessing tokens of given owner will be called on off-chain. Like TokenUri method, there is no any on-chain use case for that method.  Also as you mentioned we have gas-limit and this method may reach the gas limit easily. 

It is usually preferred design smart contract methods as O(1) way. So you can easily predict cost of method and it will not vary depending on other factors. 

If i am going to add another method which return tokens for given owner then this method is not going to compatible with current eip proposals and it will be custom.

YakupIpek:
> @noescape00 @YakupIpek can the token Id just be logged in the transactions logs somewhere on transfer? Then the receivers just need to inspect the tx Logs of the transfer to see the Id. It's not ideal but there's not really another safe way unless you do some time of pagination to get all owned tokens which can still run into gas cost issues.
> 
> There is also a FN related issue that will come into play currently if a list of anything is returned, event uint TokenIds. FN `return` value won't output array return values of any type currently.

Transactions are already logged.

noescape00:
@mrtpain 
`We can't safely do this because it would have to have some limit of how many NFTs an address could own`

If this limit is smth like ulong.MaxValue then I think we're ok

`or gas limits could prevent the user from retrieving the values if they hold too many NFTs.`

It would be a local method so gas is unlimited (unless we have a gas limit for local calls). 
Do we have gas limit for local calls and if so is there a reason behind having it other than endless loop prevention? 

@YakupIpek 
```
If i am going to add another method which return tokens for given owner then this method is not going to compatible with current eip proposals and it will be custom.
```

Why not? If a class implements an interface and adds method on top it's still compatible with any other class that implements that interface.

mrtpain:
The limits I was speaking of were gas limits and it isn't limited to local calls. If a wallet holds a ton of NFTs and wants to receive or send 1, working through and modifying an array of wallet held tokens could cause actual transactions to fail trying to transfer or receive.  @noescape00

noescape00:
Regarding the logs approach: 

Ok, if you say that's the way to go then it is. 

As far as I understand to get the logs `/api/SmartContracts/receipt-search` is used. I've tried to call it to collect logs for my contract which had maybe 6 interactions and it took 1-2 minutes to return. 

With such delays getting this info `on the fly` is unfeasible, so server-side cashing should be implemented that would collect logs for NFTs and then map owner's addresses to token ids for quick access. 

Do you know if someone is scheduled to work on this? 

@mrtpain @YakupIpek

YakupIpek:
@rowandh can this be merged ?

rowandh:
Is everyone OK with it?

YakupIpek:
@mrtpain @noescape00 ? 

@Rowan This was already reviewed actually long ago.

noescape00:
Yes, looks good.

Testnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
```diff
@@ -36,148 +34,175 @@ public struct ApprovalForAllLog
...
+
+    /// <summary>
+    /// The next token id which is going to be minted
+    /// </summary>
+    private ulong TokenIdCounter
     {
-        return GetSupportedInterfaces(interfaceID);
+        get => State.GetUInt64(nameof(TokenIdCounter));
+        set => State.SetUInt64(nameof(TokenIdCounter), value);
+    }
```
mrtpain, line 181: I think it would be beneficial to have this public. Especially if you're looking for an old TokenId you know you own.

mrtpain:
I think you probably should close these 2 open PRs and open a combined merged PR for review. 

Difficult to follow 2 PRs with many changes and give any sort of approval. 

#44 and #46 are what I'm referencing.

Testnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
YakupIpek: There are different strategies for Minting tokens. Some contracts prefer that client specify tokenid on mint method and others prefer auto incremented id numbers by contract. In our case we use auto incremented and Mint methods already returns minted tokenId. 

Client doesn't need to and have to know what is next id for contracts designed both way. 

Any application is going to rely on TokenIdCounter is going to fail when switching to another contract has no auto incremented token id strategy. 

For example, StoreContract is able to manage tokens from different ERC721 contracts and StoreContract is going to fail that if it is assumed that all these ERC721 contracts have TokenIdCounter or auto incremented token id strategy.

YakupIpek: Little bit more info Minting methods, mint/burn methods are not part of ERC721 and actually this is not issue because  applications that manage ERC721 tokens only interact with ERC721 interface so naturally mint/burn methods are private to your own application.

YakupIpek:
#46 should to be reviewed after #44 is merged so no need to so i am waiting this pr to be merged asap. 

> I think you probably should close these 2 open PRs and open a combined PR for review.
> 
> Difficult to follow 2 PRs with many changes and give any sort of approval.
> 
> #44 and #46 are what I'm referencing.

#46 should to reviewed after this pr is merged. Current pr is complete and can be reviewed & merged without #46. 

So i need this merged asap.

Testnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
```diff
@@ -36,148 +34,175 @@ public struct ApprovalForAllLog
...
     ///<summary>
     /// Gets the address of the owner of the NFT ID. 
     ///</summary>
     /// <param name="id">The ID of the NFT</param>
     ///<returns>The owner address.</returns>
-    private Address GetIdToOwner(ulong id)
-    {
-        return this.PersistentState.GetAddress(GetIdToOwnerKey(id));
-    }
+    private Address GetIdToOwner(ulong id) => State.GetAddress(GetIdToOwnerKey(id));
```
mrtpain, line 75: make public

Testnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
```diff
@@ -36,148 +34,175 @@ public struct ApprovalForAllLog
...
     /// <summary>
     /// Getting from NFT ID the approval address.
     /// </summary>
     /// <param name="id">The ID of the NFT</param>
     /// <returns>Address of the approval.</returns>
-    private Address GetIdToApproval(ulong id)
-    {
-        return this.PersistentState.GetAddress(GetIdToApprovalKey(id));
-    }
+    private Address GetIdToApproval(ulong id) => State.GetAddress(GetIdToApprovalKey(id));
```
mrtpain, line 96: public

Testnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
```diff
@@ -36,148 +34,175 @@ public struct ApprovalForAllLog
...
     /// <summary>
     /// Gets the amount of non fungible tokens the owner has.
     /// </summary>
     /// <param name="address">The address of the owner.</param>
     /// <returns>The amount of non fungible tokens.</returns>
-    private ulong GetOwnerToNFTokenCount(Address address)
-    {
-        return this.PersistentState.GetUInt64($"OwnerToNFTokenCount:{address}");
-    }
+    private ulong GetBalance(Address address) => State.GetUInt64($"Balance:{address}");
```
mrtpain, line 110: public

Testnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
```diff
@@ -36,148 +34,175 @@ public struct ApprovalForAllLog
...
     /// Gets the permission value of the operator authorization to perform actions on behalf of the owner.
     /// </summary>
     /// <param name="owner">The owner address of the NFT.</param>
     /// <param name="operatorAddress">>Address of the authorized operators</param>
     /// <returns>A value indicating if the operator has permissions to act on behalf of the owner.</returns>
-    private bool GetOwnerToOperator(Address owner, Address operatorAddress)
-    {
-        return this.PersistentState.GetBool($"OwnerToOperator:{owner}:{operatorAddress}");
-    }
+    private bool GetOwnerToOperator(Address owner, Address operatorAddress) => State.GetBool($"OwnerToOperator:{owner}:{operatorAddress}");
```
mrtpain, line 125: public

Testnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
mrtpain, line 181: There is no downside to making these getters public, I think they _should_ be public when there is no harm done. It gives flexibility for future needs that may not be thought of right now. Locking by making private makes the contract more of a black box.

mrtpain: There is no downside to making these getters public, I think they should be public when there is no harm done. It gives flexibility for future needs that may not be thought of right now. Locking by making private makes the contract more of a black box.

Testnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
```diff
@@ -467,6 +489,103 @@ private void CanTransfer(ulong tokenId)
...
+    /// <param name="tokenId">The token identifier which is being transferred.</param>
+    /// <param name="data">Additional data with no specified format.</param>
+    private void EnsureContractReceivedToken(Address from, Address to, ulong tokenId, byte[] data)
+    {
+        if (State.IsContract(to))
+        {
+            var result = Call(to, 0, "OnNonFungibleTokenReceived", new object[] { Message.Sender, from, tokenId, data }, 0);
+            Assert((bool)result.ReturnValue);
+        }
+    }
```
mrtpain, line 569: I would add a re-entrancy lock on any methods that involve this method.

Lock the contract in the beginning of any public method making an external call to any contract not directly trusted or created by this one. 
Unlock the contract after the method is finished.

Example from out contracts: 
https://github.com/Opdex/opdex-v1-core/blob/main/src/Contracts/Pools/OpdexStakingPool.cs#L192-#L207

mrtpain (commented):
It looks ok to me I think. I made more comments about public getters, Idk how the rest of the team feels, but I feel pretty strongly that anything stored in state should have public getters if there is no security risk involved. 

Give developers the power to access these values even if there isn't a direct need at the moment. Making them private can only hurt in the long run IMO. 

Ultimately up to yourself and the team 👍

Testnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
YakupIpek: These are tiny wrappers for avoid magic strings and easy access to state variables. Any required information is available as methods already. Secondly we only want to expose ERC721 related interfaces only. 

Please see OwnerOf method.

Testnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
YakupIpek: See `GetApproved(ulong tokenId)`

Testnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
YakupIpek: See `public ulong BalanceOf(Address owner)`

Testnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
YakupIpek: See public bool IsApprovedForAll(Address owner, Address operatorAddress)

Testnet/NonFungibleToken/NonFungibleToken/NonFungibleToken.cs
YakupIpek: Re-entrancy is not an attack or invalid case that should to be prevented. Re-entrancy is a perfectly valid case which should not be prevented. Methods should safely handle re-entrancy cases. 

I wouldn't recommended your lock mechanism because any exception raise between of EnsureUnlocked and UnLock methods will be locked forever. Also as i said you prevent valid use cases in this way. 

There are 2 place where EnsureContractReceivedToken called and there is no any kind of state modification after this method called so there is no security issue on contract.

YakupIpek:
@mrtpain I left my comments. Can you approve if everything is fine for you atm ?

mrtpain (approved):
I see, your comments make sense 👍 LGTM

YakupIpek:
@rowandh  It is approved by @noescape00  and @mrtpain. Can you merge this please ?

PR #47, opened 2021-08-20
Place sale token by notification mechanism

YakupIpek:


YakupIpek:
@mrtpain @noescape00 @rowandh 

It is ready for your reviews

YakupIpek:
@mrtpain Can you review this as well ?

Testnet/NFTStore/NFTStore.Tests/InMemoryState.cs
```diff
@@ -46,7 +47,8 @@ public T GetStruct<T>(string key)
 
         public UInt256 GetUInt256(string key) => GetValue<UInt256>(key);
 
-        public bool IsContract(Address address) => IsContractResult;
+        public void SetIsContract(Address address, bool isContract) => AddOrReplace($"__IsContract:{address}", isContract);
```
mrtpain, line 50: `private` right?

Testnet/NFTStore/NFTStore/NFTStore.cs
```diff
@@ -86,20 +64,40 @@ public void CancelSale(Address contract, ulong tokenId)
...
     }
 
-    private bool IsApprovedForAll(Address contract, Address tokenOwner)
+    public bool OnNonFungibleTokenReceived(Address operatorAddress, Address fromAddress, ulong tokenId, byte[] data)
     {
-        var result = Call(contract, 0, "IsApprovedForAll", new object[] { tokenOwner, Message.Sender });
+        EnsureNotPayable();
 
-        Assert(result.Success, "IsApprovedForAll method call failed.");
+        var seller = fromAddress == Address.Zero ? operatorAddress : fromAddress;
```
mrtpain, line 71: Does the `operatorAddress` need an `Addrss.Zero` check?

Testnet/NFTStore/NFTStore/NFTStore.cs
```diff
@@ -86,20 +64,40 @@ public void CancelSale(Address contract, ulong tokenId)
...
+
+        Assert(Address == GetOwner(tokenContract, tokenId), "The store contract is not owner of the token.");
+
+        var saleInfo = GetSaleInfo(tokenContract, tokenId);
+
+        Assert(saleInfo.Price == 0, "The token is already on sale.");
+                
+        SetSaleInfo(tokenContract, tokenId, new SaleInfo { Price = price, Seller = seller });
+
+        Log(new TokenOnSaleLog { Contract = tokenContract, TokenId = tokenId, Price = price, Seller = seller, Operator = operatorAddress, Order = NextId++ });
```
mrtpain, line 89: Might need to use `++NextId` so the incremented value is logged. Currently if `Next++` equals 5, 4 would be logged I believe and the State value would be incremented to 5.

I could be wrong, but you might want to double check this with tests expecting logged values.

Testnet/NFTStore/NFTStore/NFTStore.cs
```diff
@@ -173,4 +174,10 @@ public struct SaleInfo
         public ulong Price;
         public Address Seller;
     }
+
+    public struct ReceiveInfo
+    {
+        public Address Contract;
```
mrtpain, line 180: Could there be use for an index here at all?

Testnet/NFTStore/NFTStore.Tests/InMemoryState.cs
YakupIpek: No we want to able to set given address is belong to contract or not on testing stage.

YakupIpek: "__" prefix is there for avoid key conflicts

Testnet/NFTStore/NFTStore/NFTStore.cs
YakupIpek: I think it is not necessary check. There will be always an operator address actually. But fromAddress will be a zero when it is minted before this call.

Testnet/NFTStore/NFTStore/NFTStore.cs
YakupIpek: `NextId` starts from 0 and we want to start from 0 so in this case we assign it as 0 and incremented it for next usage. `++NextId` would start from 1.

Testnet/NFTStore/NFTStore/NFTStore.cs
YakupIpek: This is not a log. It is just struct saved in PersistentState

YakupIpek:
@mrtpain Hi mate i replied your reviews. Let me know if you have more questions. We want to get audited by Certik asap so we want to merge it.

Testnet/NFTStore/NFTStore/NFTStore.cs
mrtpain: Ok then my next question is why `NextId` is incrementing on both this `OnNonFungibleTokenReceived` method and on the `Buy` method. 

PRs with updates to smart contracts are a bit difficult at a glance because I can't see the full contract here, I have to go to your fork to see how some of these comments can be affected by the rest of the code.

YakupIpek: /receip-search endpoint returns logs unordered way. Client can not know order of logs when it is really important. So for this reason we provide mechanism know order of logs so every log has different id.

mrtpain: If you must do that, then you might want to use `UInt256` for the `NextId`. 

I thought the receipt-search endpoint does order transactions and their logs in the correct order? Does the receipt endpoint also return things out of order?

YakupIpek: receipt-search endpoint returns results with blocknumber so i can order results but problem is one block can contain many logs so in this case this workaround is needed.

mrtpain: Yes but aren't transactions in ascending order and their logs as well? So by a simple loop and filter, you could pick all the correct types of logs in ascending order already. Is an Id, really needed for this? 

I guess I just don't understand why you need to log and keep incrementing an NextId. 

Its not a blocker from merging, just lack of understanding on my end I think.

YakupIpek: Transactions can be ascending order in block but receip-search is different endpoint and response. It doesn't return results ordered way. Even it does so it does by only block number so this doesn't cover the case when block has many logs. 

Can you give me example how you can sort results of receip-search ?

YakupIpek: NextId++ needs a checked keyword.

mrtpain: > Transactions can be ascending order in block but receip-search is different endpoint and response. It doesn't return results ordered way. Even it does so it does by only block number so this doesn't cover the case when block has many logs.

In my experience, these logs have always been in order that I've noticed and I would expect that to be the case. If not, my opinion on this is that its a FN issue and these contracts should rely on a forever incremented `NextId++`. If for any reason this is not possible, then I think `NextId` should at least be made `UInt256` to prevent contract locks in the potential case of reaching `ulong` limits.

YakupIpek: I think you are right. Most important thing is order of logs and if they are not ordered so this should to be fixed on FN side. Otherwise receipt-search endpoint is not usable for something.

mrtpain: Another option would be receipt search to find valid transactions then retrieve the transaction again by hash to get the correct order. 

A bit more difficult.

I noticed a while back `eventType` was added to logs in this format but hides potential `eventType` properties in actual logs, spoke with @rowandh and this should be restructured in the near future. 

FROM: 
```
"logs": [
    {
        "log": {
            "event": "CreateLiquidityPoolLog",
            "token": "PHLh9aUdSBS7zke28vnsE4UiVznecFqy9y",
             "pool": "PJTHv2gp5XXhZ5SpfDcZC6qesetTNNUW47"
        }
    }
]
```

TO: 
```
"logs": [
    {
        // Can potentially add SortOrder here
        "event": "CreateLiquidityPoolLog",
        "log": {
            "token": "PHLh9aUdSBS7zke28vnsE4UiVznecFqy9y",
             "pool": "PJTHv2gp5XXhZ5SpfDcZC6qesetTNNUW47"
        }
    }
]
```

If that response is adjusted maybe logs can also be returned in their original sort order with a sortOrder property. Maybe that will help you. Just suggestions 👍

YakupIpek:
@mrtpain Ok ordering workaround removed. 

@rowandh Can you merge this pr ?

rowandh:
@YakupIpek bytecode + hash need to be updated first.

PR #50, opened 2021-10-01
[Testnet] NFT Contract - Unique URL for per Token

YakupIpek:
Uniq Url for per token.


mrtpain (commented):
So the goal of this is to allow a completely new URL per NFT? Does it matter if that URL has nothing to do with the NFT token Id? 

Otherwise LGTM but you might want to put _some_ validations on the URI in public methods. Making sure they're at the least including `http://` or `https://` preferably and not null or whitespace. Without validations makes room for accidental errors.

YakupIpek:
> So the goal of this is to allow a completely new URL per NFT? Does it matter if that URL has nothing to do with the NFT token Id?
> 
> Otherwise LGTM but you might want to put _some_ validations on the URI in public methods. Making sure they're at the least including `http://` or `https://` preferably and not null or whitespace. Without validations makes room for accidental errors.

The server or metadata may not even exist if even url parameter is perfectly valid.  Also if metadata is even exist in that moment but it may not later. So this is all client issue that has to be resolved.

Like on previous example we dont validate url on constructor and it is same in ETH implementations. Url validation may require some regex so no need to punish others expensive gas operation who pass valid urls. 

@rowandh It is ready to merge.

rowandh:
@YakupIpek could you please update the readme with the latest bytecode + hash?

PR #57, opened 2021-12-28
[Testnet] StandardTokenEnhanced contract

zeptin:
This provides a superset of the functionality in the existing `StandardToken` contract, particularly the implementation of 3 new interfaces (`IOwnable`, `IMintable`, `IBurnable`). These are roughly based on the corresponding OpenZeppelin interfaces that extend ERC20.

`IBurnableWithMetadata` is intended to provide a non-standard function that is needed for InterFlux, in particular the conversion of SRC20 tokens back into ERC20.

As this PR requires comment on the proposed implementation the bytecode has not been finalised yet.


zeptin:
@rowandh @YakupIpek @fassadlr @noescape00 @mrtpain any comments you have would be appreciated

mrtpain:
IMO these should be two separate projects entirely or functiliaty is giong to be jammed into `StandardToken` that may not be needed for all tokens. 

`StandardToken` with the basics as we have it
`StandardTokenEnhanced` with the added abilities.

We're going to have to pivot from the way ERC does things with the whitelisting hashes. Solidity contracts can be built and displayed more modular without whitelisting, while our contracts need to focus on a specific implementation per whitelisted hash.

Testnet/StandardToken/StandardToken/StandardToken.cs
```diff
@@ -163,6 +163,197 @@ public struct ApprovalLog
...
+
+    /// <inheritdoc />
+    public Address Owner
+    {
+        get => State.GetAddress(nameof(this.Owner));
+        private set => State.SetAddress(nameof(this.Owner), value);
+    }
+
+    /// <inheritdoc />
+    public void OnlyOwner()
```
mrtpain, line 266: can probably be private.

Testnet/StandardToken/StandardToken/StandardToken.cs
```diff
@@ -163,6 +163,197 @@ public struct ApprovalLog
...
+    }
+
+    /// <inheritdoc />
+    public bool IsOwner()
+    {
+        return Message.Sender == Owner;
+    }
+
+    /// <inheritdoc />
+    public void RenounceOwnership()
```
mrtpain, line 278: A certiK audit will tell you to assign a `PendingOwner`, then another method to have that `PendingOwner` come in an `ClaimOwnership` to avoid pushing ownership to a potentially invalid address.

Testnet/StandardToken/StandardToken/StandardToken.cs
```diff
@@ -163,6 +163,197 @@ public struct ApprovalLog
...
+
+        Address previousOwner = this.Owner;
+
+        this.Owner = Address.Zero;
+
+        Log(new OwnershipTransferred { PreviousOwner = previousOwner, NewOwner = Address.Zero });
+    }
+
+    /// <inheritdoc />
+    public void TransferOwnership(Address newOwner)
```
mrtpain, line 290: This does the same thing as `RenounceOwnership`?

mrtpain, line 290: A certiK audit will tell you to assign a PendingOwner, then another method to have that PendingOwner come in and ClaimOwnership to avoid pushing ownership to a potentially invalid address.

zeptin:
> IMO these should be two separate projects entirely or functionality is going to be jammed into `StandardToken` that may not be needed for all tokens.
> 
> `StandardToken` with the basics as we have it `StandardTokenEnhanced` with the added abilities.
> 
> We're going to have to pivot from the way ERC does things with the whitelisting hashes. Solidity contracts can be built and displayed more modular without whitelisting, while our contracts need to focus on a specific implementation per whitelisted hash.
> 
> If these _do_ end up being separate projects, the interfaces can be broken out into their own files and the directory of a newly `StandardTokenEnhanced` project can be compiled as long as `[Deploy]` is used, making things going forward a bit easier on the eyes too.

Yeah I think you're right, and there isn't really a great solution to the modularity issue as it stands. Obviously having one whitelisted hash for every permutation of ownable/mintable/burnable wouldn't be ideal. And that is without taking into account other possible extensions like pausable etc.

Testnet/StandardToken/StandardToken/StandardToken.cs
```diff
@@ -163,6 +163,197 @@ public struct ApprovalLog
...
+
+        return false;
+    }
+
+    /// <inheritdoc />
+    public bool BurnWithMetadata(UInt256 amount, string metadata)
+    {
+        if (Burn(amount))
+        {
+            Log(new BurnMetadata() { From = Message.Sender, Amount = amount, Metadata = metadata });
```
mrtpain, line 333: Up to you, not sure how ERC does it but the `From` and `Amount` fields will be logged within a `TransferLog` during the `Burn` => during the internal `TransferTo`.

So this log could only include the metadata, but 🤷‍♂️

mrtpain:
> Yeah I think you're right, and there isn't really a great solution to the modularity issue as it stands. Obviously having one whitelisted hash for every permutation of ownable/mintable/burnable wouldn't be ideal. And that is without taking into account other possible extensions like pausable etc.

I think overall this all looks good, I have a few minor comments, mostly only considerations. Maybe the name could be more specific but gets tough without making it too long. `StandardTokenBurnMintOwn` or `StandardTokenOwnableDynamicSupply` are a bit much but more specific.

Testnet/StandardToken/StandardToken/StandardToken.cs
zeptin: My rationale is that, if you were only searching for these metadata logs, you'd have sufficient information in them to do something actionable. And we need the amount as well as the metadata for interflux. So the duplication isn't optimal but it does simplify things elsewhere.

So it's mostly a choice of copy-pasting the transfer logic to do this specific type of burn, or have the two separate log entries with nearly identical information. I suppose having both logs does keep better compatibility with ERC20 though.

mrtpain: Thats fine too, was more of just a callout. 

I am in favor of an added `SupplyLog` though or something named similarly, so TotalSupply changes are logged for watching. Opdex could use this for sure to track token supplies as they change rather than manual check after a Mint/Burn that has `Address.Zero` logged.

noescape00 (approved):
Looks great, gj

mrtpain:
@zeptin When you're ready, if you could update the README and the description of the PR with your expected bytecode and hash, then I can review.

Testnet/StandardTokenOwnableDynamicSupply/StandardTokenOwnableDynamicSupply/StandardTokenOwnableDynamicSupplyContract.cs
```diff
@@ -0,0 +1,265 @@
...
+    public void OnlyOwner()
+    {
+        Assert(IsOwner(), "Only the owner can call this method");
+    }
+
+    /// <inheritdoc />
+    public bool IsOwner()
+    {
+        return Message.Sender == Owner;
+    }
```
drmathias, line 163: Consider making these two methods private and removing from the interface.

Testnet/StandardTokenOwnableDynamicSupply/StandardTokenOwnableDynamicSupply/StandardTokenOwnableDynamicSupplyContract.cs
```diff
@@ -0,0 +1,265 @@
...
+    /// <param name="symbol">The symbol used to identify the token.</param>
+    /// <param name="decimals">The amount of decimals for display and calculation purposes.</param>
+    public StandardTokenOwnableDynamicSupplyContract(ISmartContractState smartContractState, UInt256 totalSupply, string name, string symbol, byte decimals) : base(smartContractState)
+    {
+        this.TotalSupply = totalSupply;
+        this.Name = name;
+        this.Symbol = symbol;
+        this.Decimals = decimals;
+        this.SetBalance(Message.Sender, totalSupply);
+        this.Owner = Message.Sender;
```
mrtpain, line 25: This specifically can be problematic with the `MultisigContract` - https://github.com/stratisproject/CirrusSmartContracts/pull/59/files

The Multisig contract can't create this token contract as its currently written. Maybe adding an `owner` param to this constructor and remove the assignment using `Message.Sender`

zeptin: Just so that I'm not misunderstanding, why would it not be able to create it? Wouldn't the multisig contract address be `Message.Sender`? Or is the behaviour of sender different during creation from within another contract?

Presumably if the multisig contract were to create instances of this one directly then there would need to be a dedicated method added to the multisig for it, as the transaction submission mechanism can't handle contract creation. And eventually a method to deploy the NFT contract would need to be added as well.

Testnet/StandardToken/StandardToken/StandardToken.cs
zeptin: I have added a log for tracking supply changes

Testnet/StandardTokenOwnableDynamicSupply/StandardTokenOwnableDynamicSupply/StandardTokenOwnableDynamicSupplyContract.cs
mrtpain: The MS contract/project will need to include all code from this contract if is to be able to create these enhanced tokens and would need a method internally to call the `Create` method. Then the MS contract would properly be the owner of this contract which enables it to `Mint` tokens. 

As these are two separate contracts currently, the MS contract cannot create any contracts, so it can't be the owner of an enhanced token, so it can't mint new tokens for the supply. 

Same for NFT's then I suppose if they're assigning owners right at creation based on Message.Sender.

zeptin: Well, the way this was handled for wSTRAX is that the ERC20 was deployed by a normal EOA and then transferred to the multisig as a second transaction. This approach works on Cirrus too.

The ability for the multisig to collectively agree to create contracts by vote would be a nice to have, but as you say the project would then need every contract it would ever need to deploy inside it, and could never be extended to support additional contract types.

mrtpain: ahh I see, so you're saying this contract could be created then would use `TransferOwnership` to the multisig contract? That should be fine, wasn't considering that route.

Testnet/StandardTokenOwnableDynamicSupply/StandardTokenOwnableDynamicSupply/StandardTokenOwnableDynamicSupplyContract.cs
```diff
@@ -192,6 +188,12 @@ public void Mint(Address account, UInt256 amount)
         Log(new TransferLog() { From = Address.Zero, To = account, Amount = amount });
 
         this.TotalSupply += amount;
+
+        Log(new SupplyChangeLog()
+        {
+            PreviousSupply = (this.TotalSupply - amount),
+            TotalSupply = this.TotalSupply
+        });
```
mrtpain, line 196: Very minor but you could save some gas here with the following, same sort of comment for burn. 

``` c#
var totalSupply = this.TotalSupply;
var updatedSupply = totalSupply += amount;

TotalSupply = updatedSupply;

Log(new SupplyChangeLog()
{
    PreviousSupply = totalSupply,
    TotalSupply = updatedSupply
});

StratisIain:
Whitelisted `6402d68aee8e37cd595a2e3426dc6cdc1046fc427fb374e75c3a99596f616c03`

PR #59, opened 2022-01-05
[Testnet] Multisig Contract

zeptin:
Perhaps this 'serialised contract call metadata' functionality could be incorporated into a future version of the SmartContracts package.

This is a somewhat complex contract (particularly the `Confirm` method) so careful review is needed.

Based on the Gnosis MultiSigWallet:
https://github.com/gnosis/MultiSigWallet

### DIFF
--- a/Testnet/Multisig/Multisig/MultisigContract.cs
+++ b/Testnet/Multisig/Multisig/MultisigContract.cs
@@ -59,6 +59,11 @@ public class MultisigContract : SmartContract
         return State.GetBool($"{ConfirmationPrefix}:{transactionId}:{address}");
     }
 
+    public uint Confirmations(ulong transactionId)
+    {
+        return State.GetUInt32($"{ConfirmationCountPrefix}:{transactionId}");
+    }
+
     public Transaction GetTransaction(ulong transactionId)
     {
         Assert(transactionId <= TransactionCount, "Can't retrieve a transaction that doesn't exist yet.");
@@ -122,8 +127,10 @@ public class MultisigContract : SmartContract
     /// <summary>
     /// Submits method and parameter metadata describing a contract call that the multisig contract should invoke once it is sufficiently confirmed.
     /// </summary>
+    /// <param name="destination">The address of the contract that the multisig contract will invoke a method on.</param>
+    /// <param name="methodName">The name of the method that the multisig contract will invoke on the destination contract.</param>
     /// <param name="data">An array of method parameters encoded as packed byte arrays. See the <see cref="Transaction"/> struct for further information.</param>
-    /// <returns>The transactionId of the submitted multisig transaction.</returns>
+    /// <returns>The transactionId of the submitted multisig transaction. This is used for confirmation by other multisig nodes, and for looking up execution status etc.</returns>
     /// <remarks>The submitter implicitly provides a confirmation for the submitted transaction.</remarks>
     public ulong Submit(Address destination, string methodName, byte[] data)
     {
@@ -160,7 +167,7 @@ public class MultisigContract : SmartContract
 
         State.SetBool($"{ConfirmationPrefix}:{transactionId}:{Message.Sender}", true);
 
-        uint confirmationCount = State.GetUInt32($"{ConfirmationCountPrefix}:{transactionId}");
+        uint confirmationCount = Confirmations(transactionId);
 
         State.SetUInt32($"{ConfirmationCountPrefix}:{transactionId}", ++confirmationCount);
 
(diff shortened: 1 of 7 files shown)


Testnet/Multisig/Multisig/MultisigContract.cs
```diff
@@ -0,0 +1,355 @@
...
+        Log(new RequirementChange() { Requirement = required });
+    }
+
+    /// <summary>
+    /// In the nomenclature of the Gnosis MultisigWallet contract this is based on, the 'wallet' refers to the contract itself.
+    /// In other words, methods with this restriction are intended to only be called by the contract itself.
+    /// </summary>
+    private void EnsureWalletOnly()
+    {
+        Assert(Message.Sender == Message.ContractAddress, "Can only be called by the contract itself.");
```
mrtpain, line 291: This is clever, so it looks like transactions will be submitted by an owner, then confirmed by every other owner, then finally will call this contract from itself to execute the public method. 

Do I have that right?

I'm sure you've got it covered but this might get tricky as members join, they'll need to loop through and confirm all pending transactions. 

This also leaves a bit of a loophole, if a member confirms, then is removed. `Required` will be deducted from but the confirmations of pending transactions will remain the same. So if `Required=10` PersonA confirms Transaction. It will have 1 confirmation, then they leave and now `Required=9` but that transaction's confirmations is still 1. Somebody new can join bringing `Required=10` back but there still is that 1 confirmation but a no longer approved owner.

Testnet/Multisig/Multisig/MultisigContract.cs
```diff
@@ -0,0 +1,355 @@
...
+    /// <summary>
+    /// Called by contract owners to indicate that they accept the proposed method invocation provided via <see cref="Submit"/>.
+    /// </summary>
+    /// <param name="transactionId">The transactionId of the submitted multisig transaction.</param>
+    /// <remarks>Once the confirmation count matches or exceeds <see cref="Required"/> the contract call will be executed.</remarks>
+    public void Confirm(ulong transactionId)
+    {
+        Assert(IsOwner(Message.Sender), "Sender is not authorized to confirm.");
+        Assert(transactionId <= TransactionCount, "Transaction ID does not exist.");
+        Assert(!IsConfirmedBy(transactionId, Message.Sender), "Sender has already confirmed the transaction.");
```
drmathias, line 159: These 3 assert calls aren't necessary when this method is called by the `Submit` method. Consider splitting out into a private `ConfirmExecute` method which doesn't have these 3 asserts, as well as a public `Confirm` method, which does these asserts, then calls `ConfirmExecute`.

Also more of a nitpick, might want to use `EnsureOwnersOnly()` in place of `Assert(IsOwner(Message.Sender), ...)`

Testnet/Multisig/Multisig/MultisigContract.cs
```diff
@@ -0,0 +1,355 @@
...
+    {
+        EnsureOwnersOnly();
+
+        TransactionCount++;
+
+        State.SetStruct($"{TransactionPrefix}:{TransactionCount}", new Transaction()
+        {
+            Destination = destination,
+            Executed = false,
+            Value = 0,
```
drmathias, line 138: Am I right in thinking this only supports transactions which require not sending CRS?

Could potentially support owners paying by having a target and current transaction value, as well as a block deadline to refund failed transactions. Or slightly more simple might be to have the submitter pay.

Testnet/Multisig/Multisig/MultisigContract.cs
```diff
@@ -0,0 +1,355 @@
...
+
+        uint confirmationCount = State.GetUInt32($"{ConfirmationCountPrefix}:{transactionId}");
+
+        State.SetUInt32($"{ConfirmationCountPrefix}:{transactionId}", ++confirmationCount);
+
+        Log(new Confirmation() { Sender = Message.Sender, TransactionId = transactionId });
+
+        if (confirmationCount < Required)
+            return;
+
```
drmathias, line 171: Consider moving the code below here into its own `Execute` method. Partly for readability, though maybe would be possible to allow re-execution of failed transactions.

drmathias (commented):
Nice 👍

Testnet/Multisig/Multisig/MultisigContract.cs
zeptin: Yes, the transactions get confirmed by a configurable number of other owners. Once the final confirmation is reached the transaction is 'executed'. This currently supports arbitrary destination contracts and methods; it was tested with the enhanced standard token contract, for example.

The member removal is a bit of an issue, but the transactions are not intended to remain unconfirmed for very long under normal circumstances.

StratisIain:
Whitelisted `861b336503a9a5ce85243b14751479981ef9261e8c65bb06a9c58d233b6b135e`

PR #72, opened 2022-03-09
Fix SupplyChangeLog for burns

zeptin:


fassadlr (approved):

mrtpain (approved):
😢 yikes not sure how we missed that one....

Fix LGTM, bytecode and hash match the README files, however the PR description displays the wrong hash, should update that so if anyone looks back at this PR it has the correct hash in the details.

PR #81, opened 2022-07-13
[Testnet] Mintable Token

quantumagi:
The SC differs from the `InterfluxStandardToken` contract in the following ways:

- Uses `IPullOwnership` instead of `IOwnable`
- Adds `IMintableWithMetadata` and `MintMetadata`
- Removes `NativeChain` and `NativeAddress`
- Adds `Minter` which defaults to the `Owner` address. Adds a public `SetMinter` method.

The purpose of the SC is to support scenarios such as when tokens are minted on-demand as the result of deposits to a bank account.

I've not included the `NativeChain` and `NativeAddress` fields but perhaps some argument can be made for keeping those and setting values that would reflect a more generalized Interflux-like use-case?


Testnet/MintableToken/MintableToken/IMintableWithMetadata.cs
```diff
@@ -0,0 +1,19 @@
...
+public interface IMintableWithMetadata
+{
+    /// <summary>
+    /// The owner of the contract can create (mint) new tokens as required.
+    /// This increases the total supply of the token.
+    /// </summary>
+    /// <remarks>Emits a TransferLog event with the 'from' address set to the zero address.</remarks>
+    /// <param name="account">The account the newly minted tokens should be credited to.</param>
+    /// <param name="amount">The amount of tokens to mint.</param>
+    /// <param name="metadata">Additional data to be recorded with the burn.
```
zeptin, line 16:
```suggestion
    /// <param name="metadata">Additional data to be recorded with the mint.
```

mrtpain (approved):
LGTM

Bytecode and hash match after building the project and validating at the directory level.

StratisIain:
`63a6a3a74e2a4fd16cf223fc30d5d7c8075870091d4cc14175546a1dbe814fd8` has been whitelisted

PR #83, opened 2022-09-09
Add SubmitOrConfirm method

quantumagi:
The nodes can just call this method instead of having the onerous requirement of having to figure out whether to call `Submit` or `Confirm`. This also avoids any race conditions in that regards and thereby provides a much more robust solution.

The new process is as follows:

1. deploy multisig with m confirmations out of n signatories required
2. the target contract ownership gets transferred to the multisig contract's address
3. node A through whatever (the signatories) submit an encoded multisig contract call to the multisig contract (via `SubmitOrConfirm`) - this implicitly gives the call m confirmations out of m required
4. when a confirm call increases the number of confirmations to m, the encoded contract call is decoded by the multisig and then the multisig contract itself calls the target contract method
5. it is allowed to do this because it is the owner

The nodes can use local calls to `ulong GetTransactionId(Address destination, string methodName, byte[] parameters)` and `bool IsConfirmedBy(ulong transactionId, Address address)` to confirm their submission.


zeptin (approved):
I get the same hash and bytecode

StratisIain:
`644e450ccc4905e6717362fc9c0dbcd457df1b166fab1bd9863c6c07dc7c6034` has been whitelisted on CirrusTest

